PR #2, opened 2019-09-21
Update README with more info on JS setup

bknoles:
I set the gem up alongside a new Rails application and came across a few things that might be helpful in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,13 @@ Inertia.configure do |config|
 end
 ~~~
 
-## Configuring Axios to work with Rails CSRF protection
+## Javacript setup
+
+### Turbolinks
+
+IMPORTANT!! InertiaJS and Turbolinks do not play nicely together since they both intercept AJAX requests. If you want to use InertiaJS, you should leave Turbolinks out of your application.
+
+### CSRF tokens
 
 Under the hood, Inertia uses Axios to make `POST`, `PATCH` and `DELETE` requests. By default, Rails will not trust those requests because Axios does not grab the Rails CSRF token from the page by default.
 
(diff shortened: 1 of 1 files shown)


bknoles:
@BrandonShar want to review this README change?

rbngzlv:
Hi! This worked for me and I think that is a good addition to the readme, wiki or docs.

README.md
```diff
@@ -128,7 +128,13 @@ Inertia.configure do |config|
 end
 ~~~
 
-## Configuring Axios to work with Rails CSRF protection
+## Javacript setup
+
+### Turbolinks
+
+IMPORTANT!! InertiaJS and Turbolinks do not play nicely together since they both intercept AJAX requests. If you want to use InertiaJS, you should leave Turbolinks out of your application.
```
savroff: In my case the problem was, that we still has parts that use Turbolinks and we need it inside.
What we end up doing is:

```ruby
redirect_to happy_path, turbolinks: false
```

README.md
```diff
@@ -141,6 +147,83 @@ window.addEventListener('DOMContentLoaded', () => {
 });
 ```
 
+### Example: Using InertiaJS with React via Webpacker
```
savroff: I think we also need an example of a setup with Vue js. 
Maybe this part we should move to `Wiki` and have it per JS library

reinink:
Hey gents! I've moved all the documentation from the individual adapter readme files to the new documentation website at https://inertiajs.com. It would be awesome if you could help contribute there, especially to do the Rails docs. 👌 

Fair warning, it's a little challenging write docs for 5+ frameworks at a time. However, it can be done. 😀

README.md
bknoles: i like that better than "don't use this with turbolinks at all"... i'm added this in #5 , i'd like us to confirm it and then we'll add it to the README... thanks for letting us know!

bknoles:
@reinink noted! we'll send a PR to the Inertiajs.com repo then with updates, yea?

BrandonShar:
Closing this since the documentation has moved to a separate repo.

Issue #4, opened 2019-10-04
Accessing instance data on shared_data blocks


rbngzlv:
Hi,

First of all thanks for the rails adapter for Inertia!

I'm planing to use the `shared_data` functionality to send the current user, locale, menu items ... but I think that is not possible right now, let me explain:

If I use the examples on the docs (https://inertiajs.com/shared-data#sharing-data) they work for synchronously and partially for the lazily case, because we don't have access to the controller instance inside the block, so we can't call the url helpers, flash, current_user, methods, etc, true?

In case that this is true:

Are you using another approach to send this data to Inertia? Or are you rendering the entire layout in Rails and only the content part with Inertia (not tested) ? Is this intended or can this be an improvement?

BrandonShar:
For us personally, we're rendering things like the page header within the rails layout, but I would imagine your use case should still work.

You're saying that something like `current_user` doesn't work inside a block though? Does rails allow us to access the controller instance inside of a lambda instead of a block?

rbngzlv:
Your approach rendering the page header o similar things in the rails layout is good because they doesn't get re-rendered in subsequent requests, right?

About the "issue" here, this works:

```ruby
class ActivityCategoriesController < ApplicationController
  def index
    puts current_user # I can access
  end
end
```

Any of this works:

```ruby
class ActivityCategoriesController < ApplicationController
  inertia_share do
    { user: current_user }
  end

 inertia_share user: lambda { current_user }
end
```

They raise:

```
undefined local variable or method `current_user' for ActivityCategoriesController:Class
```

I understand that this is because we are registering the block at the level class, I will try to investigate more about this.

And thank you for helping with this.

BrandonShar:
Yes, that's exactly the pro to rendering them directly in the layout. Anything shared via `inertia_share` is going to be evaluated and sent in every inertia response. If this isn't data that's going to change from request to request, it's better to not have the extra overhead.

And thanks for the examples! That definitely makes sense (though I didn't expect it haha). I'm guessing we'll have to handle registering this function on the controller in a different way. Tagging @bknoles for some more rails knowledge.

rbngzlv:
Jaja, don't worry. I also tested one thing that worked:

This block is executed inside the controller instance, so we have access to the controller instance with `self` and I can correctly call methods like `current_user` :

[...]

PR #7, opened 2019-10-09
Execute blocks / lambdas inside controller instance

rbngzlv:
Hi,

As commented on #4, we can discuss here a possible solution to access controller instance data inside Inertia blocks.

This solutions works, but we can do a search on other gems to see how they solve this problem, if they do.

Right now my use case is:

```ruby
  inertia_share do
    {
      flash: flash.to_hash, # used for https://inertiajs.com/shared-data#flash-messages
      menu: menu_items, # custom method
      title: "Default page title", # can be overwritten calling inertia_share on child controller 
      user: current_user
    }
  end
end
```

Edit, now also supports lambdas:

```ruby
inertia_share user_count: lambda { User.count }

inertia_share general_data: {
  app_name: 'The best app',
  user: lambda { current_user },
}
```


bknoles:
hey ruben, we had to rename all the files in order to publish this on Rubygems since `inertia` already exists on rubygems...

do you want to rebase this off the updated file structure? (if not, we can create a branch and merge that one in)

rbngzlv:
@bknoles done!

PR #11, opened 2019-10-31
Add rspec helper functionality

BrandonShar:
This is a set of test helpers for rspec users. It shims in a wrapped render function by using a spy on the InertiaRenderer that allows us to grab all of the inertia data at the time of rendering while still preserving all normal functionality for integration tests. 

It can be installed in any rails project using rspec by adding:
`require 'inertia_rails/rspec'` to your `spec/rails_helper.rb` file

Here's an example spec that shows off the new functionality:

```ruby
describe '.index', inertia: true do
    it 'returns an inertia component with pending projects' do
      get :index

      expect(inertia).to have_exact_props({projects: projects, meta: 'more data'})
      expect_inertia.to include_props({projects: projects})
      expect_inertia.to render_component 'PendingProjects'
      expect_inertia.to have_exact_view_data({user: user, more_data: 'more data'})
      expect_inertia.to include_view_data({user: user})
      
      rendered_projects = inertia.props[:projects]
      rendered_user = inertia.view_data[:user]
    end
  end
``

### CODE REVIEW

File: lib/inertia_rails/rspec.rb [hunk 1]
```diff
@@ -0,0 +1,117 @@
...
+
+RSpec::Matchers.define :include_props do |expected_props|
+  match do |inertia|
+      expect(inertia.props).to include expected_props
+  end
+
+  failure_message do |inertia|
+    "expected inertia props to include #{expected_props}, instead received #{inertia.props || 'nothing'}"
+  end
+end
```
bknoles: oooh, now i get the difference between have_props and include_props.... makes sense! i thought they were equivalent at first glance. what do you think about changing it to `have_exact_props`?

lib/inertia_rails/rspec.rb
```diff
@@ -0,0 +1,117 @@
...
+  match do |inertia|
+    expect(inertia.component).to eq expected_component
+  end
+
+  failure_message do |inertia|
+    "expected rendered inertia component to be #{expected_component}, instead received #{inertia.component || 'nothing'}"
+  end
+end
+
+RSpec::Matchers.define :have_view_data do |expected_view_data|
```
bknoles: same thing here... is `have_exact_view_data` too verbose?

lib/inertia_rails/rspec.rb
BrandonShar: Yeah I definitely see what you're saying; I'm sure you wouldn't be the only person thinking that way.

Do you think it should be an outright change or an alias?

PR #13, opened 2019-10-31
fix share persisting bug

### LINKED ISSUE #8: Cannot change inertia shared data
In my local test app, if i change (or remove) shared data, the old data is still sent to the component.

i.e. setting in the controller

```
inertia_share test: 'thing'
```

making a request, then changing the controller to:

```
inertia_share test: 'another thing'
```

sends `{ "test": "thing" }` to the controller.

I believe we maybe aren't merging shared data correctly or something like that. 

Note:This also could be due to the fact I'm running the app through Docker for Mac... sometimes Docker will cache classes if you don't have the right setting in your `development.rb`. I don't think that's the case here, but noting it in case someone else picks this up.
```

### PR DESCRIPTION
**Pull request by BrandonShar**
According to the tests in #12, this should fix our persisting issue by clearing out the shared data after rendering.

Let me know if this seems crazy.

fixes #8

### CODE REVIEW

**Comment by bknoles** [comment]
very clever! reset seems like the right idea, but i wonder if we could do it on the beginning or end of an inertia request instead?

i'm wondering if something like this:

```ruby
inertia_share time: Time.now.to_s

def some_inertia_action
  if some_condition
    render :inertia
  else
    redirect_to "/a_non_inertia_route"
  end
end
```

would print the stale `time` value on the next request after `some_condition` evaluates to false.

what about resetting in the middleware? that runs before the request gets to rails doesn't it?

BrandonShar:
Great point! I originally wanted it to do it before, but couldn't think of the right place to put it. I think the middleware makes sense. Seems like it should happen pretty much first thing, regardless of whether or not it's an inertia request. Sound right?

PR #17, opened 2019-11-24
Rewrite "inertia_share" to drop `InertiaRails.reset!`

### LINKED ISSUE #15: inertia_share fails on second request
I'm playing around with `inertia_rails` within a Rails 6 application like this:

```ruby
  inertia_share do
    {
      foo: 42
    }
  end
end
```

This works fine on the first request (props include `foo:42`), but all subsequent requests fail (props not include `foo:42`). When I modify my source file (`application_controller.rb`) and change `foo` to 43, it works on the following one request, but the next requests will fail again.

This can be reproduced with the test suite when you swap the order of the two context blocks in this file: https://github.com/inertiajs/inertia-rails/blob/a27564bb512e6bdb28517e8b2b4de39419a57371/spec/inertia/sharing_spec.rb#L4-L18

Seems to be caused by #13.

ledermann:
Fixing #15 needs some rewrite. Why this is important, please see the added test in `spec/inertia/sharing_spec.rb`: Because of `InertiaRails.reset!`, all shared data is lost on second request of the same URL. 

My rewrite of `inertia_share` makes the added test pass, without touching the existing tests (except for small details) or changing the API.

Background: Managing shared data via `mattr_accessor` of the `InertiaRails` module does not allow to use class inheritance at controller level, e.g. defining some global shared data in `ApplicationController` and some additional in `UsersController` but not in `SomeOtherController`. This can't be achieved by managing them in **one** global array. In addition, it can't handle code changes which was described in #8.

What the rewrite does:

- It uses `class_attribute` placed in the base controller, so every controller can define it's own shared data. Some code from the module `InertiaRails` is moved over.
- It removes `InertiaRails.reset!`, because it's not needed anymore

This fixes #15 and is a better fix for #8. It reverts the changes from #13.

### DIFF
--- a/lib/inertia_rails/controller.rb
+++ b/lib/inertia_rails/controller.rb
@@ -5,19 +5,46 @@ module InertiaRails
     extend ActiveSupport::Concern
 
     included do
-      before_action do
+      error_sharing = proc do
         # :inertia_errors are deleted from the session by the middleware
-        InertiaRails.share(errors: session[:inertia_errors]) if session[:inertia_errors].present?
+        if @_request && session[:inertia_errors].present?
+          { errors: session[:inertia_errors] }
+        else
+          {}
+        end
+      end
+
+      if Gem::Version.new(Rails::VERSION::STRING) >= Gem::Version.new('5.2.0')
+        class_attribute :shared_plain_data, default: {}
+        class_attribute :shared_blocks, default: [error_sharing]
+      else
+        # In older Rails there is no `default` for class_attribute. This was
+        # introduced with Rails 5.2: https://github.com/rails/rails/pull/29270
+        class_attribute :shared_plain_data
+        class_attribute :shared_blocks
+
+        self.shared_plain_data = {}
+        self.shared_blocks = [error_sharing]
       end
     end
 
-    module ClassMethods
-      def inertia_share(**args, &block)
-        before_action do
-          InertiaRails.share(**args) if args
-          InertiaRails.share_block(block) if block
-        end
+    class_methods do
+      def inertia_share(hash = nil, &block)
+        share_plain_data(hash) if hash
+        share_block(&block) if block_given?
       end
+
+      def share_plain_data(hash)
+        self.shared_plain_data = shared_plain_data.merge(hash)
+      end
+
+      def share_block(&block)
+        self.shared_blocks = shared_blocks + [ block ]
+      end
+    end
+
+    def shared_data
+      shared_plain_data.merge(evaluated_blocks)
     end
 
     def redirect_to(options = {}, response_options = {})
@@ -45,6 +72,12 @@ module InertiaRails
       if (inertia_errors = options.dig(:inertia, :errors))
         session[:inertia_errors] = inertia_errors
       end
+    end
+
+    private
+
+    def evaluated_blocks
+      shared_blocks.map { |block| instance_exec(&block) }.reduce(&:merge) || {}
     end
   end
 end
--- a/lib/inertia_rails/inertia_rails.rb
+++ b/lib/inertia_rails/inertia_rails.rb
@@ -3,16 +3,8 @@ require 'active_support/core_ext/module/attribute_accessors_per_thread'
 require 'inertia_rails/lazy'
 
 module InertiaRails
-  thread_mattr_accessor :threadsafe_shared_plain_data
-  thread_mattr_accessor :threadsafe_shared_blocks
-
   def self.configure
     yield(Configuration)
-  end
-
-  # "Getters"
-  def self.shared_data(controller)
-    shared_plain_data.merge!(evaluated_blocks(controller, shared_blocks))
   end
 
   def self.version
@@ -21,20 +13,6 @@ module InertiaRails
 
   def self.layout
     Configuration.layout
-  end
-
-  # "Setters"
-  def self.share(**args)
-    self.shared_plain_data = self.shared_plain_data.merge(args)
-  end
-
-  def self.share_block(block)
-    self.shared_blocks = self.shared_blocks + [block]
-  end
-
-  def self.reset!
-    self.shared_plain_data = {}
-    self.shared_blocks = []
   end
 
   def self.lazy(value = nil, &block)
@@ -67,9 +45,5 @@ module InertiaRails
 
   def self.shared_blocks=(val)
     self.threadsafe_shared_blocks = val
-  end
-
-  def self.evaluated_blocks(controller,  blocks)
-    blocks.flat_map { |block| controller.instance_exec(&block) }.reduce(&:merge) || {}
   end
 end
--- a/lib/inertia_rails/middleware.rb
+++ b/lib/inertia_rails/middleware.rb
@@ -19,9 +19,7 @@ module InertiaRails
         status, headers, body = @app.call(@env)
         request = ActionDispatch::Request.new(@env)
 
-        ::InertiaRails.reset!
-
-        # Inertia errors are added to the session via redirect_to 
+        # Inertia errors are added to the session via redirect_to
         request.session.delete(:inertia_errors) unless keep_inertia_errors?(status)
 
         status = 303 if inertia_non_post_redirect?(status)
@@ -92,4 +90,3 @@ module InertiaRails
     end
   end
 end
-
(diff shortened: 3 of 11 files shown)


BrandonShar:
Thanks for the PR @ledermann ! 

To make sure I understand the root cause, when you hit the same route twice, is rails is doing something like re-using the controller instance so it doesn't run the `inertia_share` method a second time?

bknoles:
same question as brandon on the missing data on second request. I'm not sure why `inertia_share` wouldn't work twice in a row with the original code... `reset!` should run before the request and `inertia_share` would run again once the controller loads, no? is there some kinda of caching happening that we haven't considered?

bknoles:
Regarding the refactor of the `inertia_share` methods, making the inertia_share data per controller instance instead of global is an interesting idea.

If you look at the inertia-laravel docs:
https://inertiajs.com/shared-data

> The server-side adapters provide a way to preassign shared data for each request. This is typically done outside of your controllers.

you can see that the original design intent was for a single application wide shared data store.

we kinda strayed from that pattern by adding the inertia_share as a controller method... though i think we kind of had to... brandon is there a Rails equivalent to the `AppServiceProvider` used in the Laravel example?

With that context in mind, I would say that the refactor has the advantage of making `inertia_share` more flexible with the disadvantage of diverging from the patterns of `inertia-laravel`... do you guys agree with that assessment?

bknoles:
also, thanks for writing this up @ledermann !

ledermann:
Thank you both for looking into my PR! Let me explain my proposed code change more in detail:

With the original code, `inertia_share` doesn't work. Only the first time, when the Ruby code (where you placed `inertia_share`) is loaded, the two vars `shared_plain_data` and `shared_blocks` are set and used, but then the middleware strikes and resets both vars, which will never be set again. This is because code in the class body is run only once: When the class is loaded.

Try this in your application:

```ruby
  Rails.logger.info 'The ApplicationController class is loaded.'

  inertia_share do
    Rails.logger.info 'inertia_share is called.'

    { foo: 42 }
  end
end
```

At runtime, both info messages are logged **one** time - when the class is loaded. When the second request comes in, Ruby uses the code from the memory and doesn't read the file again and so does not execute code at the class body again. So the two vars are empty for all upcoming requests.

In development mode there is a little difference: If you change a Ruby file, it will be reloaded so the methods in the class body are executed again - but this will help for just the next request - after this, the vars will be reset again.

Therefore a fix is absolutely necessary.

Regarding the inertia-laravel docs ("This is typically done outside of your controllers"): Using Ruby on Rails I think this should be the `ApplicationController` which is the base class of all controllers. Placing code into `ApplicationController` **is** the Rails way of an application wide shared data store which is available to all controllers of the app. You can't use this code from outside the controller context (e.g. from the models), but I wouldn't recommend that either - `inertia_share` is about adding generic data to responses, so the controller context is the right place for doing this, IMHO.

bknoles:
@ledermann great explanation... makes a lot of sense. i'm not sure why we assumed `inertia_share` would run on every request... perhaps too much time spent in the Laravel world where every file executes fresh on every request.

I've created #19 , which should address the issue without as much refactoring. The test you added passes in that PR as well.

The change may also behave in a similar fashion as your refactor, which stores the data at the controller class level instead of the module level. I added a quick test with an inherited controller, and it passes. However, I haven't thought through whether redirects would cause issues with either this PR or #19.

Would you mind taking a look at #19 and letting me know your thoughts on it?

ledermann:
#19 seems to be ok, but IMHO the shared data is still being stored in the wrong place. 

I've added two additional tests to my PR, one for the inherited controller and one for redirects. If anything is still missing or can be improved, I will gladly do it as well.

ledermann:
@bknoles In the meantime I found another issue with the current implementation (see #20), so I would like to propose again for this PR to merge.

BrandonShar:
Thanks for the extra example @ledermann , we'll take another look at this soon!

bknoles:
I haven't forgotten about this... I have been thinking it over here and there. :)

After poking around in this branch a bit, I see two main differences in this method for sharing data from what is currently in the gem.

1. In this branch, shared data is local to the controller rendering Inertia. In the current gem, shared data is scoped to the top level `InertiaRails` module.
2. In this branch, shared data is loaded once, the first time the controller class is loaded. In the current gem, shared data is cleared before each request, and added to the module variable in a `before_action` hook. In both cases, shared blocks and any lambdas inside plain data are evaluated fresh on every request during `render inertia: 'BlahComponent'`

I do not have a strong opinion on the first point. Functionally, these will behave the same in almost all use cases, I believe. The current version follows the design of the original Laravel package, which has a single store of shared data... I do think that is a slight benefit. The version here has a benefit of only running `inertia_share` the first time a class is loaded and cached... Georg has identified another difference in #20, but I actually prefer the current behavior that makes shared data available to a controller catching the exception [note: there is a bug in the current behavior that needs to be addressed, if possible].

On the second point, I do have a preference for resetting shared data between each request. In #8, we had an issue with seeing stale data in development. We fixed that, and this branch doesn't have that behavior (I pulled it down and checked in a dummy app); but because we use module/class variables to store data, I think there is some risk that we could end up with stale data in some case. I tried and failed to create a test which would share stale data, so I am fairly confident that we are safe... but I am _extremely_ confident that making Inertia share explicitly stateless between requests will prevent that issue.

In the end, I don't see a clear use case that would drive me to want to make this change. There definitely might be one, though, so my "vote" goes for leaving this out for now and keeping eyes open for a place where module variable vs. local controller variables makes a difference.

@BrandonShar could you weigh in as well? Specifically, on #20, as that is a place where the behavior _is_ different. Would you want to see shared data in that ExceptionsController? Or would you prefer to  have the shared data cleared out there?

ledermann:
@bknoles Thank you very much for your detailed thinking about my PR. I'm happy to get this great feedback! I see your points, but I still vote for my PR ;-)

Some short arguments for merging:

1) The main cause for using`inertia_share` is to **define** something, not to **do** something. So, for me this definition should not be run for every single request. Maybe performance or memory is not a real issue here, be IMHO this extra work is not required.

2) IMHO, shared data should not be seen as global stuff. Think of multiple controllers and inherited controllers, all defining there own shared data. Stuffing it all into the same bucket seems wrong to me. Emptying the bucket on every request to get rid of stuff from other controllers is a kind of hack.

3) My PR fixes the bug described in https://github.com/inertiajs/inertia-rails/issues/20#issuecomment-564676880. There is no other solution in sight.

4) It don't think that the stale issue (#8) can occur again, because this bug was caused by having a global storage which was not reset by reloading the controller. But exactly this is fixed by this PR, the storage in class attributes at the controller level is reset by reloading the controller.

No matter what you decide: I am grateful that the Rails adapter exists and would like to continue to contribute in the future - if you are interested :-)

bknoles:
> No matter what you decide: I am grateful that the Rails adapter exists and would like to continue to contribute in the future - if you are interested :-)

@ledermann absolutely!! thanks for all your help on the gem so far. i personally appreciate your thoroughness, and it is really cool to see other people interested in it.

BrandonShar:
Hey folks, I'm going to close this PR because I believe all of the issues it raised have been solved in other ways.

If that's not the case, please feel free to open it back up with an explanation!

ElMassimo:
TL'DR: `InertiaRails.reset!` should be called [inside an `ensure` block](https://github.com/ElMassimo/request_store_rails/blob/master/lib/request_store_rails/middleware.rb#L18-L19), otherwise unhandled errors can __leak data and blocks__  across unrelated requests in unrelated controllers.

---

I can confirm that some of the bugs described here still exist, and that while https://github.com/inertiajs/inertia-rails/pull/38 removed some of the race conditions, it still hasn't addressed the problem where a fatal error halts the middleware chain preventing `InertiaRails.reset!` to be called.

As a result, any subsequent request in the same thread will have `shared_data` and `shared_blocks` from the previous request, which is incorrect and dangerous (can cause requests to fail, and data to be leaked across requests).

I'll put together a proper bug report, but wanted to mention that the refactor added here by @ledermann has clear advantages when working with multi-threaded servers like Puma.

Shared global state should be avoided like the plague, and I say this [from experience](https://github.com/ElMassimo/request_store_rails) 😄

PR #18, opened 2019-11-24
Run tests with GitHub Actions

ledermann:
The use of GitHub Action is very simple. There was an incomplete Travis file that doesn't seem to be in use, so I removed it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 /pkg/
 /spec/reports/
 /tmp/
+/Gemfile.lock
 
 /spec/dummy/db/*.sqlite3
 /spec/dummy/db/*.sqlite3-journal
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -1,0 +1,26 @@
+name: Testing
+
+on: [push]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: Setup System
+        run: sudo apt-get install libsqlite3-dev
+
+      - name: Set up Ruby 2.6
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -6,6 +6,7 @@
 /pkg/
 /spec/reports/
 /tmp/
+/Gemfile.lock
```
bknoles: i wasn't sure about this, but this comment:
https://github.com/radar/guides/issues/14#issuecomment-328864865

succinctly describes why gemfile.lock shouldn't be used for gems:

> Since the gem command only respects the dependency version constraints named in the gemspec, not Gemfile.lock, the presence of a Gemfile.lock is more likely to keep gem maintainers unaware of incompatibility with newer versions of dependencies that gem users will install but maintainers will not.

bknoles:
what will running the tests in Github Actions do for us? I'm assuming it will just alert everyone if we accidentally merge in a failing change? (this is a genuine question, I haven't dug into Github Actions at all)

.gitignore
ledermann: Exactly. The most referenced article about this is https://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/

ledermann:
Using a CI (by using GitHub Actions or Travis) helps to ensure that code changes don't break the tests. The tests are run for every commit, which is very helpful for PRs: Without checking out the code of the contributor, you can see if the tests are passing. GitHub Actions is free for public repos and very simple to set up, so IMHO these are low hanging fruits.

On more thing: In GitHub settings of the repo you can define branch protections and make sure that only code changes with green tests can be merged in:
https://help.github.com/en/github/administering-a-repository/about-required-status-checks

PR #19, opened 2019-11-27
Fix condition where inertia_share data gets erased for forever on second request

### LINKED ISSUE #15: inertia_share fails on second request

bknoles:
Closes #15 

As @ledermann points out in #15 and PR #17 , the calls to `inertia_share` from a controller only run the first time the class is loaded and the data is subsequently deleted forever on the next request, because of the `InertiaRails.reset!` call in the middleware... presumably, we didn't notice this in development either due to classes reloading when changed, or maybe because `config.cache_classes = false` in development environments.

The PR does a lot of refactoring that we want to discuss before merging in, so I created this PR, which solves the bug without changing much code.

The basic idea is to wrap the `InertiaRails.share` calls inside a block given to `before_action`. This, combined with the current `InertiaRails.reset!` call in the middleware, ensures that shared data is recalculated from scratch on each request.

I wrote a test to check inheritance as well, which seems to work as well, though this change isn't explicitly intended to add support for inheritance.

### DIFF
--- a/lib/inertia_rails/controller.rb
+++ b/lib/inertia_rails/controller.rb
@@ -6,8 +6,10 @@ module InertiaRails
 
     module ClassMethods
       def inertia_share(**args, &block)
-        InertiaRails.share(args) if args
-        InertiaRails.share_block(block) if block
+        before_action do
+          InertiaRails.share(args) if args
+          InertiaRails.share_block(block) if block
+        end
       end
     end
   end
--- a/spec/inertia/sharing_spec.rb
+++ b/spec/inertia/sharing_spec.rb
@@ -1,19 +1,40 @@
 RSpec.describe 'using inertia share when rendering views', type: :request do
   subject { JSON.parse(response.body)['props'].symbolize_keys }
-  
+
   context 'using inertia share' do
     let(:props) { {name: 'Brandon', sport: 'hockey', position: 'center', number: 29} }
     before { get share_path, headers: {'X-Inertia' => true} }
-    
+
     it { is_expected.to eq props }
   end
 
-  context 'inertia share across requests' do 
+  context 'inertia share across requests' do
     before do
-      get share_path, headers: {'X-Inertia' => true} 
+      get share_path, headers: {'X-Inertia' => true}
       get empty_test_path, headers: {'X-Inertia' => true}
     end
-  
+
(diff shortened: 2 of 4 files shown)


lib/inertia_rails/controller.rb
```diff
@@ -6,8 +6,10 @@ module Controller
 
     module ClassMethods
       def inertia_share(**args, &block)
-        InertiaRails.share(args) if args
-        InertiaRails.share_block(block) if block
+        before_action do
```
bknoles: because we clear data in the middleware, we need to re-share the data on each request

spec/inertia/sharing_spec.rb
```diff
@@ -1,19 +1,40 @@
...
+
+    before do
+      get share_path, headers: {'X-Inertia' => true}
+      get share_path, headers: {'X-Inertia' => true}
+    end
+
+    it { is_expected.to eq props }
+  end
+
+  context 'using inertia share with inheritance' do
```
bknoles: this seems to work with inherited controllers, though there may be some other cases which behave differently

ledermann:
Adding and deleting the shared data on every single request doesn't seem right to me. But ok, this fixes the issue with minimal code changes :-)

ledermann:
Any news here? Without merging a patch (either this PR or #17) the gem is broken.

BrandonShar:
Thanks for the bump @ledermann ! It's been a busy week for us.

bknoles:
yes, my apologies... last week was a holiday here and cutting the new gem version fell off my todo list! thanks for picking it up brandon!

Issue #22, opened 2020-01-15
Middleware fails with Rack v2.1.0/2.1.1


ledermann:
Recently, the gem `rack` was released as [v2.1.0](https://github.com/rack/rack/blob/master/CHANGELOG.md#210---2020-01-10) and [v2.1.1](https://github.com/rack/rack/blob/master/CHANGELOG.md#211---2020-01-12) with lots of changes. 

These updates are not compatible with inertial-rails - the test suite fails:

```
Failures:

  1) InertiaRails::Middleware the version is stale tells the client to refresh
     Failure/Error: get empty_test_path, headers: {'X-Inertia' => true, 'X-Inertia-Version' => 'blkajdf'}

     NoMethodError:
       undefined method `[]' for nil:NilClass
     # /Users/ledermann/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/rack-2.1.1/lib/rack/etag.rb:38:in `call'
     # /Users/ledermann/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/rack-2.1.1/lib/rack/conditional_get.rb:27:in `call'
     # /Users/ledermann/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/rack-2.1.1/lib/rack/head.rb:14:in `call'
     # /Users/ledermann/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/actionpack-6.0.2.1/lib/action_dispatch/http/content_security_policy.rb:18:in `call'
     # /Users/ledermann/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/rack-2.1.1/lib/rack/session/abstract/id.rb:277:in `context'
     # /Users/ledermann/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/rack-2.1.1/lib/rack/session/abstract/id.rb:271:in `call'
     # /Users/ledermann/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/actionpack-6.0.2.1/lib/action_dispatch/middleware/cookies.rb:648:in `call'
     # /Users/ledermann/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/actionpack-6.0.2.1/lib/action_dispatch/middleware/callbacks.rb:27:in `block in call'
[...]

PR #23, opened 2020-01-22
Fix breaking rack change

BrandonShar:
As @ledermann identified in #22 , Rack released a breaking change in a patch release.

This commit https://github.com/rack/rack/commit/72959ebc2f300f3b2ccb7ae2aae9f199e611dfb6#diff-10b933d2c1fdc82ceecade456c64e1c2 removed the implicit cast to array `to_ary` method from `Rack::Response`, which we were unintentionally relying on.

I switched to using the `finish` method as the rack::response code comments suggest. It's an alias to `to_a` that has been in the codebase for over a decade, so I feel pretty safe :smile:

### DIFF
--- a/lib/inertia_rails/middleware.rb
+++ b/lib/inertia_rails/middleware.rb
@@ -40,7 +40,7 @@ module InertiaRails
   
     def force_refresh(request)
       request.flash.keep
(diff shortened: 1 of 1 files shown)


BrandonShar:
Fixes #22

PR #24, opened 2020-01-23
Add an inertia? method onto request

BrandonShar:
I've been using this in my own project and noticed that @reinink added it to the laravel version. 

So why not send it upstream!

`request.inertia?` is an easy way to check in your controller whether or not you're dealing with an inertia request. In my project, I use it to decide whether to return an http status or an inertia response when I'm handling things like auto-save.

### DIFF
--- a/lib/inertia_rails.rb
+++ b/lib/inertia_rails.rb
@@ -3,6 +3,7 @@ require 'inertia_rails/engine'
 
 require 'patches/debug_exceptions'
 require 'patches/better_errors'
+require 'patches/request'
 
 ActionController::Renderers.add :inertia do |component, options|
   InertiaRails::Renderer.new(
--- a/lib/patches/request.rb
+++ b/lib/patches/request.rb
@@ -1,7 +1,5 @@
-if defined?(ActionDispatch::Request)
-  ActionDispatch::Request.class_eval do
-    def inertia?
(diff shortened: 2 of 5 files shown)


BrandonShar:
@ledermann would you mind reviewing this to make sure I'm using monkey patching best practices? I'm fairly new to it in ruby.

ledermann:
Pretty nice extension!

The MonkeyPatch looks good to me, maybe you could alternatively use `prepend`:

```ruby
# lib/patches/request.rb
module InertiaRequest
  def inertia?
    key? 'HTTP_X_INERTIA'
  end
end

ActionDispatch::Request.prepend InertiaRequest
```

IMHO, checking that `ActionDispatch::Request` is defined seems not required, because in other places of the gem this class is already used. To be on the safe side, a runtime dependency should be added to `inertia_rails.gemspec`:

```ruby
spec.add_dependency 'actionpack'
```

BrandonShar:
Good call on not needing to check for `ActionDispatch`, we'd have bigger problems elsewhere if it wasn't present haha.  And I switched to that `key?` syntax since that's really nice looking.

Does `prepend` make a difference here since we're creating a new method instead of overriding an existing one?

ledermann:
I agree, using `prepend` will be the same as `include` here, because there is no override. Using `class_eval` is also completely fine.

PR #25, opened 2020-02-18
Fix typo (intertia_rails) in README.md

larssg:
* Changes intertia_rails to inertia_rails in README.md


bknoles:
good catch! you have no idea how many times i've typed "intertia" with two "t"s into my editor... my fingers are convinced that is how it's actually spelled and i can't change their mind!

PR #30, opened 2020-05-05
Add base path configuration option

smlparry:
Hey there 👋 

I ran into an issue with Inertia that I think could be fixed quite easily. I'm using nginx to reverse proxy my subdomain to a nested route in Rails, the configuration looks like this:
Screen Shot 2020-05-05 at 11 26 26 am

Which causes Inertia to rewrite the current path to `/admin` rather than just `/`

inertia-base-path

Ive managed to fix this by setting a before action in my controller as such:

Screen Shot 2020-05-05 at 11 27 43 am

And this works correctly:
inertia-base-path-fixed

It seemed easy enough to add this as a configuration option to the Rails gem and I would imagine it being set in the initializer

FYI I havn't tested this code, but I hope it illustrates the point :)

### DIFF
--- a/lib/inertia_rails/inertia_rails.rb
+++ b/lib/inertia_rails/inertia_rails.rb
@@ -19,6 +19,10 @@ module InertiaRails
     Configuration.layout
   end
 
+  def self.base_path
+    Configuration.base_path
+  end
+
   # "Setters"
   def self.share(**args)
     shared_plain_data.merge!(args)
@@ -38,6 +42,7 @@ module InertiaRails
   module Configuration
     mattr_accessor(:layout) { 'application' }
     mattr_accessor(:version) { nil }
+    mattr_accessor(:base_path) { nil }
 
     def self.evaluated_version
       self.version.respond_to?(:call) ? self.version.call : self.version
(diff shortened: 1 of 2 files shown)


bknoles:
Hi @smlparry thanks for opening up the PR here...

Let me see if i understand this correctly...

You are using nginx to send browser traffic to `trainer.fitterapp.local` to the `trainer.fitterapp.local/admin` route within the rails app. 

Is this so you can nest admin based routes/resources in the /admin namespace within Rails?

And then also you don't want the user to see `/admin` in the url?

smlparry:
Thats exactly right :)

bknoles:
i see... wouldn't this condition still exist without using inertia? for example, a theoretical update action with standard, non-inertia Rails code:

```ruby
def update
  @thing = Thing.find(params[:id])
  if @thing.update(thing_params)
    redirect_to admin_thing_path(@thing)
  else
    render :edit
  end
end
```

will redirect the user to `http://trainer.fitterapp.local/admin/things/:thing_id` right?

BrandonShar:
Closing because we don't think this is something inertia should be responsible for given that the same shortcomings exist in rails. Feel free to re-open (or add an issue) if we're mistaken!

PR #31, opened 2020-05-19
Fix Ruby 2.7 deprecation warnings

ledermann:
When using Ruby 2.7, the gem displays same deprecation warnings:

```
lib/inertia_rails/rspec.rb:64: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
...
lib/inertia_rails/controller.rb:10: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
```

I fixed them and enabled matrix testing to ensure I didn't break anything for Ruby 2.6.

I updated some development dependencies as well - this reduces warnings while running tests.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -12,10 +12,10 @@ jobs:
       - name: Setup System
         run: sudo apt-get install libsqlite3-dev
 
-      - name: Set up Ruby 2.6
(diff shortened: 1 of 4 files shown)


BrandonShar:
Thanks @ledermann !

PR #33, opened 2020-06-01
Add an inertia_partial? method onto request

ledermann:
Similar to #24 this PR adds `#inertia_partial?` to the request. It checks the header for `X-Inertia-Partial-Data` to distinguish between a full or partial request.

Example use case: In an application showing posts, I want to Increase the post's view counter for full requests only. Code from the controller:

```ruby
  after_action :increase_view_count, only: [ :show ]
  # ...

  private

  def increase_view_count
    @post.punch(request) unless request.inertia_partial?
  end
end
```

### DIFF
--- a/lib/patches/request.rb
+++ b/lib/patches/request.rb
@@ -2,4 +2,8 @@ ActionDispatch::Request.class_eval do
   def inertia?
     key? 'HTTP_X_INERTIA'
   end
+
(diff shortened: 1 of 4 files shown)


BrandonShar:
I haven't used this feature of inertia yet, but this makes good sense to me. Thanks!

PR #34, opened 2020-06-01
Test pull requests with GitHub Actions, too

ledermann:
Currently, the tests are run for pushes only, not for pull requests. I fixed this, so we can see the checks for an incoming PR.


BrandonShar:
At some point, I'm going to end up copy+pasting this file into other projects. Thanks for setting this up and maintaining it!

PR #36, opened 2020-07-01
Fix spec.homepage in inertia_rails.gemspec

crispincornett:
Remove trailing slash from `spec.homepage` in inertia_rails.gemspec, which led to an invalid link for the `changelog_uri`


BrandonShar:
Thanks @crispincornett !

PR #38, opened 2020-07-07
Make inertia_share threadsafe

### LINKED ISSUE #35: inertia_share doesn't always seem to run
We're using Inertia for a new web app and loving things so far. We do currently have an issue where the inertia_share doesn't seem to run consistently on initial page load?

The rendered props are there and correct in the `data-page` attribute but the shared props are nowhere to be seen. A reload typically fixes this but it's pretty bad in our instance since we share a `layout` prop which renders an application shell

bknoles:
Resolves #35 

`mattr_accessor` is not thread safe... this can cause rendered props to be incorrect through the following process:

1. request 1 hits server
2. The middleware runs `Inertia.reset!` and clears shared data
3. `inertia_share` stores data in the `InertiaRails` module variables
4. request 2 hits server to a _different controller_
5. The middleware runs `Inertia.reset!` and clears shared data
6. request 1 renders data to the view. since shared data has been cleared, the shared data from step 3 is not included in the response to request 1

My knowledge of threading in Rails goes as deep as the ~50 articles/docs pages that i read about it today. This solution assumes that a single thread only processes a single request at a time.

### DIFF
--- a/lib/inertia_rails/inertia_rails.rb
+++ b/lib/inertia_rails/inertia_rails.rb
@@ -1,6 +1,6 @@
 module InertiaRails
-  thread_mattr_accessor :shared_plain_data
-  thread_mattr_accessor :shared_blocks
+  thread_mattr_accessor :threadsafe_shared_plain_data
+  thread_mattr_accessor :threadsafe_shared_blocks
 
   def self.configure
     yield(Configuration)
@@ -8,7 +8,7 @@ module InertiaRails
 
   # "Getters"
   def self.shared_data(controller)
-    threadsafe_shared_plain_data.merge!(evaluated_blocks(controller, threadsafe_shared_blocks))
+    shared_plain_data.merge!(evaluated_blocks(controller, shared_blocks))
   end
 
   def self.version
@@ -21,11 +21,11 @@ module InertiaRails
 
   # "Setters"
   def self.share(**args)
-    self.shared_plain_data = threadsafe_shared_plain_data.merge(args)
+    self.shared_plain_data = self.shared_plain_data.merge(args)
   end
 
   def self.share_block(block)
-    self.shared_blocks = threadsafe_shared_blocks + [block]
+    self.shared_blocks = self.shared_blocks + [block]
   end
 
   def self.reset!
@@ -44,20 +44,24 @@ module InertiaRails
     end
   end
 
+  # Getters and setters to provide default values for the threadsafe attributes
+  def self.shared_plain_data
+    self.threadsafe_shared_plain_data || {}
+  end
+
+  def self.shared_plain_data=(val)
+    self.threadsafe_shared_plain_data = val
+  end
+
+  def self.shared_blocks
+    self.threadsafe_shared_blocks || []
+  end
+
+  def self.shared_blocks=(val)
+    self.threadsafe_shared_blocks = val
+  end
+
   def self.evaluated_blocks(controller,  blocks)
     blocks.flat_map { |block| controller.instance_exec(&block) }.reduce(&:merge) || {}
-  end
-
-  # thread_mattr_accessor doesn't accept :default as an option, and since the method
-  # is actually defined in the context of Thread.current, we can't do something like:
-  # def self.shared_plain_data
-  #   @@shared_plain_data || {}
-  # end
-  def self.threadsafe_shared_plain_data
-    self.shared_plain_data || {}
-  end
-
-  def self.threadsafe_shared_blocks
-    self.shared_blocks || []
   end
 end
--- a/lib/inertia_rails/inertia_rails.rb
+++ b/lib/inertia_rails/inertia_rails.rb
@@ -36,11 +36,19 @@ module InertiaRails
   private
 
   module Configuration
-    mattr_accessor(:layout) { 'application' }
+    thread_mattr_accessor :threadsafe_layout
     mattr_accessor(:version) { nil }
 
     def self.evaluated_version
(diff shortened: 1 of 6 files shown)


lib/inertia_rails/inertia_rails.rb
```diff
@@ -47,4 +47,17 @@ def self.evaluated_version
...
+  # def self.shared_plain_data
+  #   @@shared_plain_data || {}
+  # end
+  def self.threadsafe_shared_plain_data
+    self.shared_plain_data || {}
+  end
+
+  def self.threadsafe_shared_blocks
+    self.shared_blocks || []
+  end
```
bknoles, line 62: this syntax is a bit awkward... it's not as easy to define a default value for the the thread safe `mattr_accessor`

spec/inertia/sharing_spec.rb
```diff
@@ -37,4 +37,40 @@
...
+        # controller with different values for inertia_share
+        InertiaRails.reset!
+        InertiaRails.share(name: 'Brian', has_goat_status: false)
+      end
+
+      # Thread 1 starts. The controller method runs inertia_share, then sleeps.
+      # Thread 2 then modifies the shared inertia data before Thread 1 stops sleeping
+      start_thread1 = true
+      sleep 0.5
+      start_thread2 = true
```
bknoles, line 69: it's a bit clunky to get the threads to execute in the proper order to reproduce the bug, but it worked as I expected it to

snaptopixel:
Thanks for digging into this @bknoles

lib/inertia_rails/inertia_rails.rb
BrandonShar: I think this design makes sense overall, but would it be clearer of an external api to add getters and setters and treat the internal value as private? something like

```ruby
thread_mattr_accessor :threadsafe_shared_plain_data

def self.shared_plain_data
   self.threadsafe_shared_plain_data || {}
end

def self.shared_plain_data=(val)
   self.threadsafe_shared_plain_data = val
end
```

spec/inertia/sharing_spec.rb
```diff
@@ -37,4 +37,40 @@
...
+        sleep 0.1 unless start_thread1
+
+        get share_multithreaded_path, headers: {'X-Inertia' => true}
+        expect(subject).to eq props
+      end
+
+      thread2 = Thread.new do
+        sleep 0.1 unless start_thread2
+
+        # Would prefer to make this a second get request, but RSpec will overwrite
```
BrandonShar, line 57: good comment!

bknoles:
@BrandonShar changed the variable names to match your suggestion. I also realized we need thread safety for the Inertia configuration, since it can be modified during a request.

one more check please?

bknoles:
haha @ledermann i don't take it as grumbling at all... your point is well made, and it helps me to understand your perspective. Reasonable people can disagree on things, and we welcome the discussion! 

The analogy to `has_many` makes a lot of sense.  To me, since Inertia share data just boils down to a plain hash that gets passed to the rendered component, they aren't exactly equivalent. But inertia share methods/values get cached on the controller classes in production, and that _has_ caused a couple bugs already, so I'm holding my own opinion loosely.

This would be an easier discussion for me if Inertia were Rails-only or even Rails-first. In fact, the Laravel adapter was the first one created, and we are trying to copy it as well as we can.

image

Replicating what is in the image is difficult in Rails... I believe the closest equivalent to the AppServiceProvider in Rails would be a config initializer file. However, I wouldn't feel good about code like this (which should work since we pass the controller instance to the evaluated blocks):

```ruby
# config/initializers/my_inertia_share.rb
inertia_share do
  {
    user: current_user
  }
end
```

The controller is, as far as I can tell, the only logical place to define Inertia share within Rails. Since this very explicitly diverges from the Laravel design, I worry about further diverging in a way that makes the functionality completely different. Sticking to a single store for the shared data is an attempt keep us "tethered" to the Laravel adapter's design. We may find a bug that makes it impossible to straddle both designs, but until we find it, this is the rationale for keeping what is already in place.

Thanks for taking for taking the time to write out your thoughts on this, and for all your contributions to the gem... it was very cool to watch Github Actions test the gem against multiple Ruby versions when I opened this PR :)

ledermann:
Ok, because I'm a Rails guy without any experience in Laravel and PHP, I didn't thought about differences between the adapters. But from my point of view, it's more important to support the target framework as best as possible than to build adapters which are similar across the frameworks.

I agree with you that in the Rails world the controller is the right place to define shared data for Inertia. Using an initializer seems wrong to me, because that requires to pass the controller instance (as you described).

I'm using my fork (including #17) in my side project without any issues. Can you tell me about bugs with this approach? I would be happy to make my PR more robust, if there are any issues.

snaptopixel:
@ledermann @bknoles enjoying the back and forth... you guys are great communicators 🤓 interestingly enough, the discussion here about keeping with Laravel relates to the same concern in the frontend @reinink brought up here https://github.com/inertiajs/inertia/issues/145

bknoles:
@ledermann i wasn't clear there... i meant to say that the approach that is currently in the gem (a single module level store) has already had a couple bugs that we've had to patch. the first patch was adding `reset!` and the second patch was making sure the store is thread safe.

for that reason, i'm keeping my opinion open on switching to using a store per controller. i don't think we've seen any bugs related to your approach.

bknoles:
@snaptopixel thanks for sharing that link! it is super helpful to see how other parts of the community think about these things

PR #40, opened 2020-07-10
Fix using thread_mattr_accessor

### LINKED ISSUE #39: NoMethodError while starting app with gem v1.4.0
Using `inertia_rails` v1.4.0 can abort application start with this error message (caused by #38):

```
NoMethodError: undefined method `thread_mattr_accessor' for InertiaRails:Module
```

Example here:
https://github.com/ledermann/pingcrm/pull/235/checks?check_run_id=856003834

Not every Rails application is affected. If an application uses `require "rails/all"` (see [The Rails Initialization Process](https://guides.rubyonrails.org/initialization.html#loading-rails)), all works fine, because `ActionText` requires this module [explicitly](https://github.com/rails/rails/blob/v6.0.3.2/actiontext/lib/action_text/content.rb#L3). But if the frameworks are picked selectively (leaving out `ActionText`), the module may be missing and the startup may fail.

There is a workaround by adding this line to `application.rb` of the application:

```ruby
require "active_support/core_ext/module/attribute_accessors_per_thread"
```

As a fix, this `require` statement should be added to the gem.

ledermann:
This PR first strips down the dummy application to a minimalistic one: It removes all Rails frameworks which are not required to use Inertia (e.g. Sprockets, ActionText etc.)

This will cause the test suite to **fail**, because `thread_mattr_accessor` is not loaded anymore. In a second commit, the missing core extension from ActiveSupport will be loaded explicitly, so the tests are successful again.

Fixes #39


ledermann:
Because `thread_mattr_accessor` requires Rails 5 or later, I've changed the runtime dependency and removed some patches for Rails 4.

BrandonShar:
Thanks @ledermann ! This all makes sense to me. I'll let @bknoles check as well, but we should be able to get this merged and tagged today.

bknoles:
yep looks good to me too... thanks @ledermann !

PR #41, opened 2020-07-10
Test against multiple Rails versions

ledermann:
This PR enables testing against multiple Rails versions (6.0, 5.2, 5.1 and 5.0) using the [appraisal](https://github.com/thoughtbot/appraisal) gem. It creates one check for each combination of Ruby version and Rails version.


ledermann:
Tests with Rails 5.0 and 5.1 are currently failing because of the ActiveStorage config in the dummy application. Because #39 will remove this stuff, I will rebase this branch after #39 is merged.

PR #42, opened 2020-07-13
Fix error handling by returning original HTTP status

ledermann:
Currently, an InertiaRails response [has always the status code 200](https://github.com/inertiajs/inertia-rails/blob/v1.5.0/lib/inertia_rails/renderer.rb#L21). This seems wrong to me, because if there was an error (like 404 or 500), the corresponding status code should be returned, not 200.

The Laravel adapter [does the same thing](https://github.com/inertiajs/inertia-laravel/blob/v0.2.15/src/Response.php#L81), but the docs recommends [overriding this behavior in the application](https://inertiajs.com/error-handling#production) - from what I understand.

This thing was [discussed for the Laravel adapter](https://github.com/inertiajs/inertia-laravel/issues/56) some time ago. In Rails, error handling is different than in Laravel, so in my opinion returning the original status code should be returned. This fix allows [handling errors I am doing in the PingCRM demo application](https://github.com/ledermann/pingcrm/blob/master/config/application.rb#L34-L41).

Demo using the PingCRM for Rails demo application:

* Go to https://pingcrm.ledermann.dev and login
* Open developer console and open the Network tab
* Klick on the "404 error" button => The error is returned with code 200, which is **wrong**
* Open https://pingcrm.ledermann.dev/error-404 directly in the address line or do a full page reload, so a non-Inertia-request is made => Now the error is returned with code 404, which is **correct**

### DIFF
--- a/lib/inertia_rails/renderer.rb
+++ b/lib/inertia_rails/renderer.rb
@@ -18,7 +18,7 @@ module InertiaRails
       if @request.headers['X-Inertia']
         @response.set_header('Vary', 'Accept')
         @response.set_header('X-Inertia', 'true')
-        @render_method.call json: page, status: 200
+        @render_method.call json: page, status: @response.status
       else
         @render_method.call template: 'inertia', layout: ::InertiaRails.layout, locals: (view_data).merge({page: page})
       end
--- a/spec/dummy/app/controllers/inertia_test_controller.rb
+++ b/spec/dummy/app/controllers/inertia_test_controller.rb
@@ -23,11 +23,18 @@ class InertiaTestController < ApplicationController
     end
   end
 
-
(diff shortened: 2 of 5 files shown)


ledermann:
Hi @bknoles and @BrandonShar, what do you think? IMHO this is a simple fix for an obvious bug.

bknoles:
@ledermann does this conflict with #50 at all (conceptually)?

@reinink could you weigh in on whether returning `404`s (or similar statuses) instead of `200`s for error statuses fits into Inertialand ok?

ledermann:
@bknoles I don't see any conflicts with #50. Also, the tests are green after merging both branches together. No issues from my side.

PR #44, opened 2020-07-19
Set content_type for JSON response

ledermann:
Currently, a JSON response from `inertia-rails` does not always have the right content-type. See this example:

```ruby
  def index
    respond_to do |format|
      format.xml
      format.html do
        render inertia: 'Posts/Index', props: {
          ......
        }
      end
    end
  end
```

An Inertia request sets the header `Accept: text/html, application/xhtml+xml` and `X-Inertia: true` - but it gets JSON from the backend. If using the above code, the `content_type` (or `media_type`) of this JSON response is wrong: It is set to `text/html` instead of `application/json`. **This is caused by using `respond_to` !**

This PR fixes this, so Inertia is compatible with `respond_to`. Please see #42 for another fix of the JSON response.

### DIFF
--- a/lib/inertia_rails/renderer.rb
+++ b/lib/inertia_rails/renderer.rb
@@ -18,7 +18,7 @@ module InertiaRails
       if @request.headers['X-Inertia']
         @response.set_header('Vary', 'Accept')
         @response.set_header('X-Inertia', 'true')
-        @render_method.call json: page, status: 200
+        @render_method.call json: page, status: 200, content_type: Mime[:json]
       else
         @render_method.call template: 'inertia', layout: ::InertiaRails.layout, locals: (view_data).merge({page: page})
       end
--- a/spec/dummy/app/controllers/inertia_test_controller.rb
+++ b/spec/dummy/app/controllers/inertia_test_controller.rb
@@ -23,11 +23,17 @@ class InertiaTestController < ApplicationController
     end
   end
 
-
   # Calling it my_location to avoid this in Rails 5.0
   # https://github.com/rails/rails/issues/28033
   def my_location
     puts "Got to location for some reason?"
     inertia_location empty_test_path
+  end
+
+  def content_type_test
+    respond_to do |format|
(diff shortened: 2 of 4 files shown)


BrandonShar:
Can you explain a bit about your use case for this? Wouldn't an inertia route typically only return content meant to be consumed by inertia rather than something like xml?

ledermann:
Using `respond_to` in a Rails controller sends exactly the content type the browser has requested. When the browser requests `text/html` (via the `Accept` header), the controller sends `text/html`. It's pretty standard.

But Inertia works a bit different: It sends requests [always with](https://github.com/inertiajs/inertia/blob/v0.1.9/src/inertia.js#L65) `Accept: text/html`. But normally it really expects a JSON response. This conflicts with `respond_to` in Rails, because Rails sends JSON but with `text/html` content type.

Use case example: If a controller action should handle not only Inertia response, but other requests too (e.g. ATOM feed), using `respond_to` is the Rails way to handle this. But this way the controller send's the wrong content_type for Inertia requests: It sends `text/html` instead of `application/json`. 

My PR fixes this. In the current state of the gem, `respond_to` cannot to be used to answer Inertia requests.

BrandonShar (approved):
:+1:

(better late than never approval 😄 )

choznerol:
Another use case may be when used with gems like [`snaky_camel`][1]. This gem transforms JSON request body to snake_case and JSON response body to camelCase automatically [if](https://github.com/p4sh4/snaky_camel/blob/master/lib/snaky_camel/middleware.rb#L18) `Content-Type` is `application/json`.

With this PR merge, uses of [`snaky_camel`][1] can write snake_case hash key in their Rails controller...
```ruby
render inertia: 'MyApp', props: foo_bar: 42
```

...and then access it via camelCase key in js
```jsx
const MyApp = (props) => <h1>{ props.fooBar }</h1>
```

[1]: https://github.com/p4sh4/snaky_camel

PR #46, opened 2020-08-05
Configuration: Fix thread handling

### LINKED ISSUE #45: BUG: Asset versioning broken since v1.4.0
#38 introduced a serious bug within `InertiaRails.version`: It returns `nil` for every server response, so asset versioning is broken, which means the frontend will not be notified if the assets have changed.

Why? To me it seems to be caused by the thread-safe change. Now there is `threadsafe_version`, so changing the version in one thread is not shared with other threads. This seems to be intended, because there is an [explicit test for this](https://github.com/inertiajs/inertia-rails/blob/v1.4.0/spec/inertia/configuration_spec.rb#L41-L71).

IMHO the thread handling is not correct. Typically, the version is set just one time within an initializer. Using a multi threaded application server like [Puma](https://puma.io/) spawns some new threads. So far as I know, spawning a new thread does **not** run the initializers again, so the `Thread.current` does not include the `thread_safe_version` for all running threads.

Try this is out in a Rails application using inertia_rails v1.4.0:

```ruby
rails c
(irb)> InertiaRails.version.inspect
=> "1.0"
(irb)> thread = Thread.new { p InertiaRails.version.inspect }
"nil"
```

IMHO, the `layout` config has the same issue, but because it [has a fallback](https://github.com/inertiajs/inertia-rails/blob/v1.4.0/lib/inertia_rails/inertia_rails.rb#L47), it will never be `nil`.

ledermann:
This fixes #45 by reverting parts of #38. IMHO, `layout` and `version` are global and **must** be shared across threads.

To change `layout` or `version` at runtime, maybe we need another approach, e.g. by setting parameters in the `render :inertia`. Using `InertiaRails.configure` within a controller action seems not right to me.

### DIFF
--- a/lib/inertia_rails/inertia_rails.rb
+++ b/lib/inertia_rails/inertia_rails.rb
@@ -39,19 +39,11 @@ module InertiaRails
   private
 
   module Configuration
-    thread_mattr_accessor :threadsafe_layout
+    mattr_accessor(:layout) { 'application' }
     mattr_accessor(:version) { nil }
 
     def self.evaluated_version
       self.version.respond_to?(:call) ? self.version.call : self.version
-    end
-
-    def self.layout
-      self.threadsafe_layout || 'application'
-    end
-
-    def self.layout=(val)
-      self.threadsafe_layout = val
     end
   end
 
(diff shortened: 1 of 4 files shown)


BrandonShar:
Thanks @ledermann!

PR #50, opened 2020-10-27
Built in error sharing in redirects

### LINKED ISSUE #47: Automatically share validation errors
Hey there folks! Thanks so much for your ongoing maintenance of this adapter. ❤️ 

I recently updated the Laravel adapter to automatically share validation errors. You can read more about that in the release notes [here](https://github.com/inertiajs/inertia-laravel/releases/tag/v0.2.9).

Would it makes sense do add this kind of functionality to the Rails adapter as well? 🤔

bknoles:
Resolves #47 

The above issue and https://github.com/inertiajs/inertiajs.com/pull/31 both deal with validation error handling in Rails.

## Background

Initially, as noted in #47, my instinct was to try to handle error states without a redirect, as you would do within a vanilla Rails app. The discussion in https://github.com/inertiajs/inertiajs.com/pull/31 made it clear that real world apps are using the session to store errors across redirects.

I played around in my local environment with trying to implement the non-redirect pattern, which roughly looks like this in the controller:

```ruby
if contact.save
  redirect_to contacts_path, notice: 'Contact created.'
else
  render inertia: 'Users/new', {
    contact: contact,
    errors: contact.errors,
  }
end
```

I got _really close_ to making it work, but I couldn't find a satisfying UX for what happens if you submit a form with a stale version of inertia (to trigger assets reloading).

So, it makes sense to move forward with the solution in https://github.com/inertiajs/inertiajs.com/pull/31 as the "official" technique for the Rails library.

This PR copies in the work from @ledermann into the gem, with a few tweaks to make things work with `409` conflict situations.

## Notes

@ledermann suggested overwriting `redirect_to` to improve the API. I was hesitant to override the method globally, so instead I created an `inertia_redirect_to` method instead, which does the same thing that Georg wrote out.

Dealing with stale versions is tricky. I wanted to do exactly what Georg did, which was delete the errors from the session when they get shared into the Inertia store. Unfortunately, the Inertia sharing runs even when there is an assets conflict resulting in a `409`. That makes it impossible for the middleware to forward the errors from the session onto the subsequent responses in a `PATCH` -> `redirect` -> `GET` -> `409` -> re-`GET` pattern (this is the basic pattern that happens when a user submits a form with a stale Inertia version). I ended up moving the code that deletes the inertia errors into the middleware, instead of the controller. That way the middleware can keep the errors around during a stale version situation. I don't love that the single feature is split into two places, so any suggestions are welcome there!

The final API from the Rails side looks like this:

```ruby
# things_controller.rb
def new
  thing = Thing.new
  render inertia: 'NewThingComponent`, props: { thing: thing }
end

def create
  thing = Thing.new(thing_params)
  if thing.save
    # a regular redirect_to will work here also
    inertia_redirect_to thing_path(thing)
  else
    inertia_redirect_to new_thing_path, errors: thing.errors
  end
end
```

If a user tries to create a new `thing` with bad data, the `NewThingComponent` will get rendered and receive an `errors` prop. Thanks to Inertia (pointed out by both @ledermann and @BilalBudhani), the malformed data will still be around in the form.

If the user does the same thing, but with a stale version, the `things/new` page will hard refresh (due to the assets conflict). The submitted data is lost, but the errors will remain. The user would see a blank form that at least has helpful information like `Thing's name cannot be blank!` (assuming the `NewThingComponent` prints out the errors it receives). It would be really cool if the actual model data could persist across the hard refresh, but I think that might be out of the scope of the library. In any case, since this is probably an infrequent event for most applications, my opinion is that this UX is acceptable.

Thanks to @reinink and @ledermann for leading the way on this (and for being patient!)

I would love feedback here, as the middleware and error handling is a bit of a juggling act. There are specs testing it, and I actually spun up a small dummy app to verify the behavior, but I would not be surprised if there's a use case or unintended consequence that I missed.

### DIFF
--- a/lib/inertia_rails/controller.rb
+++ b/lib/inertia_rails/controller.rb
@@ -25,12 +25,12 @@ module InertiaRails
       head :conflict
     end
 
-    def inertia_redirect_to(options = {}, response_options = {})
-      if (errors = response_options.delete(:errors))
-        session[:inertia_errors] = errors
+    def redirect_to(options = {}, response_options = {})
+      if (inertia_errors = response_options.fetch(:inertia, {}).fetch(:errors, nil))
+        session[:inertia_errors] = inertia_errors
       end
 
-      redirect_to(options, response_options)
+      super(options, response_options)
     end
   end
 end
--- a/lib/inertia_rails/controller.rb
+++ b/lib/inertia_rails/controller.rb
@@ -4,6 +4,13 @@ module InertiaRails
   module Controller
     extend ActiveSupport::Concern
 
+    included do
+      before_action do
+        # :inertia_errors are deleted from the session by the middleware
+        InertiaRails.share(errors: session[:inertia_errors]) if session[:inertia_errors].present?
+      end
+    end
+
     module ClassMethods
       def inertia_share(**args, &block)
         before_action do
@@ -16,6 +23,14 @@ module InertiaRails
     def inertia_location(url)
       headers['X-Inertia-Location'] = url
       head :conflict
+    end
+
+    def inertia_redirect_to(options = {}, response_options = {})
+      if (errors = response_options.delete(:errors))
+        session[:inertia_errors] = errors
+      end
+
+      redirect_to(options, response_options)
     end
   end
 end
--- a/lib/inertia_rails/middleware.rb
+++ b/lib/inertia_rails/middleware.rb
@@ -3,41 +3,78 @@ module InertiaRails
     def initialize(app)
       @app = app
     end
-  
+
     def call(env)
+      @env = env
+
       status, headers, body = @app.call(env)
       request = ActionDispatch::Request.new(env)
 
       ::InertiaRails.reset!
-  
-      return [status, headers, body] unless env['HTTP_X_INERTIA'].present?
-  
-      return force_refresh(request) if stale?(env['REQUEST_METHOD'], env['HTTP_X_INERTIA_VERSION'])
-  
-      if is_redirect_status?(status) &&
-          is_non_get_redirectable_method?(env['REQUEST_METHOD'])
-        status = 303
-      end
-  
-      [status, headers, body]
+
+      # Inertia errors are added to the session via inertia_redirect_to 
+      request.session.delete(:inertia_errors) unless keep_inertia_errors?(status)
+
+      status = 303 if inertia_non_post_redirect?(status)
+
+      return stale_inertia_get? ? force_refresh(request) : [status, headers, body]
     end
-  
+
     private
-  
-    def is_redirect_status?(status)
+
+    def keep_inertia_errors?(status)
+      redirect_status?(status) || stale_inertia_request?
+    end
+
+    def stale_inertia_request?
+      inertia_request? && version_stale?
+    end
+
+    def redirect_status?(status)
       [301, 302].include? status
     end
-  
-    def is_non_get_redirectable_method?(request_method)
+
+    def non_get_redirectable_method?
       ['PUT', 'PATCH', 'DELETE'].include? request_method
     end
-  
-    def stale?(request_method, inertia_version)
-      sent_version = InertiaRails.version.is_a?(Numeric) ? inertia_version.to_f : inertia_version
-      saved_version = InertiaRails.version.is_a?(Numeric) ? InertiaRails.version.to_f : InertiaRails.version
-      request_method == 'GET' && sent_version != saved_version
+
+    def inertia_non_post_redirect?(status)
+      inertia_request? && redirect_status?(status) && non_get_redirectable_method?
     end
-  
+
+    def stale_inertia_get?
+      get? && stale_inertia_request?
+    end
+
+    def get?
+      request_method == 'GET'
+    end
+
+    def request_method
+      @env['REQUEST_METHOD']
+    end
+
+    def inertia_version
+      @env['HTTP_X_INERTIA_VERSION']
+    end
+
+    def inertia_request?
+      @env['HTTP_X_INERTIA'].present?
+    end
+
+    def version_stale?
+      sent_version != saved_version
+    end
+
+    def sent_version
+      return nil if inertia_version.nil?
+      InertiaRails.version.is_a?(Numeric) ? inertia_version.to_f : inertia_version
+    end
+
+    def saved_version
+      InertiaRails.version.is_a?(Numeric) ? InertiaRails.version.to_f : InertiaRails.version
+    end
+
     def force_refresh(request)
       request.flash.keep
       Rack::Response.new('', 409, {'X-Inertia-Location' => request.original_url}).finish
--- a/lib/inertia_rails/middleware.rb
+++ b/lib/inertia_rails/middleware.rb
@@ -12,7 +12,7 @@ module InertiaRails
 
       ::InertiaRails.reset!
 
-      # Inertia errors are added to the session via inertia_redirect_to 
+      # Inertia errors are added to the session via redirect_to 
       request.session.delete(:inertia_errors) unless keep_inertia_errors?(status)
 
       status = 303 if inertia_non_post_redirect?(status)
--- a/spec/dummy/app/controllers/inertia_test_controller.rb
+++ b/spec/dummy/app/controllers/inertia_test_controller.rb
@@ -27,7 +27,6 @@ class InertiaTestController < ApplicationController
   # Calling it my_location to avoid this in Rails 5.0
   # https://github.com/rails/rails/issues/28033
   def my_location
-    puts "Got to location for some reason?"
     inertia_location empty_test_path
   end
 
--- a/spec/dummy/app/controllers/inertia_test_controller.rb
+++ b/spec/dummy/app/controllers/inertia_test_controller.rb
@@ -30,11 +30,7 @@ class InertiaTestController < ApplicationController
     inertia_location empty_test_path
   end
 
-  def regular_inertia_redirect_to
-    inertia_redirect_to empty_test_path
-  end
-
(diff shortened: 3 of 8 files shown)


lib/inertia_rails/controller.rb
```diff
@@ -4,6 +4,13 @@ module InertiaRails
   module Controller
     extend ActiveSupport::Concern
 
+    included do
+      before_action do
+        # :inertia_errors are deleted from the session by the middleware
+        InertiaRails.share(errors: session[:inertia_errors]) if session[:inertia_errors].present?
```
bknoles, line 10: here we automatically share the `session[:inertia_errors]` into an `errors` prop

lib/inertia_rails/controller.rb
```diff
@@ -17,5 +24,13 @@ def inertia_location(url)
       headers['X-Inertia-Location'] = url
       head :conflict
     end
+
+    def inertia_redirect_to(options = {}, response_options = {})
+      if (errors = response_options.delete(:errors))
+        session[:inertia_errors] = errors
```
bknoles, line 30: in the session, we use the `:inertia_errors` key so that we (hopefully) don't collide with the including application's session variables

lib/inertia_rails/middleware.rb
```diff
@@ -3,41 +3,78 @@ class Middleware
...
+      return stale_inertia_get? ? force_refresh(request) : [status, headers, body]
+    end
+
     private
-  
-    def is_redirect_status?(status)
+
+    def keep_inertia_errors?(status)
+      redirect_status?(status) || stale_inertia_request?
+    end
```
bknoles, line 27: I _think_ this is the right logic for when to keep errors across requests, but i'm only 80% confident on that

lib/inertia_rails/middleware.rb
```diff
@@ -3,41 +3,78 @@ class Middleware
...
-  
+
+      # Inertia errors are added to the session via inertia_redirect_to 
+      request.session.delete(:inertia_errors) unless keep_inertia_errors?(status)
+
+      status = 303 if inertia_non_post_redirect?(status)
+
+      return stale_inertia_get? ? force_refresh(request) : [status, headers, body]
+    end
+
```
bknoles, line 22: I always had trouble digesting what we initially wrote here with the multiple return statements, so I refactored things with

1. a lot more (hopefully descriptive) helper methods
2. a single return statement

spec/inertia/error_sharing_spec.rb
```diff
@@ -0,0 +1,37 @@
+RSpec.describe 'errors shared automatically', type: :request do
```
bknoles, line 1: This is intended to test the whole "sharing errors across requests" workflow.

spec/inertia/response_spec.rb
```diff
@@ -1,10 +1,42 @@
-RSpec.describe 'Inertia::Response', type: :request do
+RSpec.describe 'InertiaRails::Response', type: :request do
```
bknoles, line 1: every PR has to have at least one pedantic thing, right? 😆

spec/inertia/response_spec.rb
```diff
@@ -1,10 +1,42 @@
...
+      it 'behaves like a regular redirect_to' do
+        post regular_inertia_redirect_to_path
+        expect(response.status).to eq 302
+        expect(response.headers['Location']).to eq(empty_test_url)
+        expect(session[:inertia_errors]).not_to be
+      end
+    end
+
+    context 'with an :errors option' do
+      # In practice, a GET -> redirect + errors probably shouldn't happen
```
bknoles, line 22: I was on the fence on whether to forward the session errors on a get request. It would be an atypical pattern in the controller, but I figured it was better to make the behavior consistent for different request types

spec/dummy/app/controllers/inertia_test_controller.rb
```diff
@@ -27,7 +27,14 @@ def inertia_partial_request_test
   # Calling it my_location to avoid this in Rails 5.0
   # https://github.com/rails/rails/issues/28033
   def my_location
-    puts "Got to location for some reason?"
```
bknoles, line 30: hey how'd you sneak in here?

lib/inertia_rails/controller.rb
BrandonShar: What is the collision risk here? It looks like we're deleting the session errors anyway so why not use that location?

spec/inertia/error_sharing_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      # Simulate that the POST was using a stale version
+      get empty_test_path, headers: headers.merge({ 'X-Inertia-Version' => 'stale' })
+      expect(response.status).to eq(409)
+      # Inertia errors are _not_ deleted
+      expect(session[:inertia_errors]).to include({ uh: 'oh' }.as_json)
+
+      # Simulate the page refresh that Inertia triggers in response to a 409
+      get empty_test_path
+      expect(response.body).to include(CGI::escape_html({ errors: { uh: 'oh' } }.to_json))
+      expect(session[:inertia_errors]).not_to be
```
BrandonShar, line 34: This is way too philosophically deep of an assertion 😂

lib/inertia_rails/controller.rb
```diff
@@ -17,5 +24,13 @@ def inertia_location(url)
       headers['X-Inertia-Location'] = url
       head :conflict
     end
+
+    def inertia_redirect_to(options = {}, response_options = {})
```
BrandonShar, line 28: I don't love the separate method name here because it feels a bit like a gotcha.. I can picture people wondering why errors aren't working and then having to say "oh yeah, you have to use inertia's special redirect_to". 

That being said I also share your unease about overwriting rail's `redirect_to` from a package..

No solution here, just thinking out loud.

lib/inertia_rails/middleware.rb
```diff
@@ -3,41 +3,78 @@ class Middleware
...
+    def stale_inertia_get?
+      get? && stale_inertia_request?
     end
-  
+
+    def get?
+      request_method == 'GET'
+    end
+
+    def request_method
```
BrandonShar, line 53: I like these new methods; super clear to read. Since they now depend on the `@env` instance variable being set from within the `call` method, do you think we could put them under a `protected` scope to avoid calling them without that initialization?

BrandonShar (approved):
This makes good sense to me and seems to line up with what other rail's users expect; good stuff!

Just a couple minor comments.

bknoles:
@BrandonShar  I can't respond directly to your first comment for some reason... but re:

> What is the collision risk here? It looks like we're deleting the session errors anyway so why not use that location?

If someone happens to be using `session[:errors]` for something else in their application, we'd run into a conflict if we try to stuff the inertia errors into that same key. the `:inertia_errors` key is only used internally. the user never sees it. they only see:

```ruby
inertia_redirect_to some_path, errors: @thing.errors
```

and then in the javascript:

```javascript
const MyComponent = ({ thing, errors }) => {
  return (
    {errors && (
      <p>Thanks inertia_rails for sending me the errors!</p>
    )}
  );
};
```

lib/inertia_rails/controller.rb
bknoles: would you feel better about:

```ruby
redirect_to some_path, inertia_errors: { watch: 'out' }
```

or maybe

```ruby
redirect_to some_path, inertia: { errors: { watch: 'out' } }
```

the latter would give us flexibility to add more options in the future at the expense of being a bit more verbose.

i don't have a problem overriding `redirect_to` per se, just not with a generic key like `errors`. i want to scope this to inertia, but also in a way that will be intuitive for people

lib/inertia_rails/middleware.rb
bknoles: to be honest, i really don't know... i believe initialization of the middleware happens when the server starts, and the only public method is `call` which is called through the the magic of the framework. i'm not sure how we could run into trouble using `@env` here, but i'm not a middleware guru by any stretch. is there a specific scenario you have in mind? or more that protected would just be following the best practice here?

lib/inertia_rails/controller.rb
BrandonShar: I would actually, both of those seem really nice. No strong opinion between the two, but I probably lean towards the second since it feels like it adds an entire "inertia" section.

lib/inertia_rails/middleware.rb
BrandonShar: Yeah, definitely not middleware specific, I was just thinking of this object as a plain class. Since we can instantiate the object without an `env`, I thought it would be good to ensure those methods would never be called outside of the `call` method. But, your point is solid that this is a middleware specific class that isn't meant to be instantiated anywhere else.

bknoles: i agree it feels weird to see instance variables being set outside of `initialize`

lib/inertia_rails/controller.rb
bknoles: cool, i will make that change

dpflucas:
Hi @reinink @bknoles - I see that this feature was already merged but isn't yet documented: https://inertiajs.com/validation#sharing-errors

Any reason for that? I'm happy to create a PR for that if the issue is simply no one did it already :)

Thanks for the fantastic work, I've been using Inertia with Rails for a while now and I'm loving it!

BrandonShar:
Hey @dpflucas creating a PR for the docs would be awesome!

I believe the change would need to be made here: https://github.com/inertiajs/inertiajs.com/blob/master/pages/validation.mdx

dpflucas:
@BrandonShar Added a simple example in https://github.com/inertiajs/inertiajs.com/pull/177 :)

PR #52, opened 2020-11-21
Lazy props

### LINKED ISSUE #51: Add lazy props feature
So I just added "lazy props" to the Laravel adapter: https://github.com/inertiajs/inertia-laravel/pull/175. It would be awesome to bring this same behaviour to the Rails adapter.

Lazy props are simply data that's *not* included on the initial visit to a page. It's only ever included when doing partial reloads, using the `only` key.

It was actually very easy to implement in the Laravel adapter. I created a new `LazyProps` class, that acts a "marker" for these props. Then, on regular (non partial reload) visits, I simply filter these props out.

bknoles:
Resolves #51 

The code follows the design of the Laravel adapter pretty closely. The basic idea is to filter out any "lazy" props from initial page loads.

There's a new `InertiaRails::Lazy` class that acts like a callable prop.

I also took the opportunity to refactor the code that decides which props get filtered out within the renderer.

### DIFF
--- a/lib/inertia_rails/lazy.rb
+++ b/lib/inertia_rails/lazy.rb
@@ -1,0 +1,23 @@
+class InertiaRails::Lazy
+  def initialize(value = nil, &block)
+    @value = value
+    @block = block
+  end
+
+  def call
+    to_proc.call
+  end
+
+  def to_proc
+    # copy to local variables so they are in scope for controller.instance_exec
+    value = @value
+    block = @block
+    if value.respond_to?(:call)
+      value
+    elsif value
+      Proc.new { value }
+    else
+      block
+    end
+  end
+end
--- a/lib/inertia_rails/lazy.rb
+++ b/lib/inertia_rails/lazy.rb
@@ -9,7 +9,10 @@ class InertiaRails::Lazy
   end
 
   def to_proc
-    # copy to local variables so they are in scope for controller.instance_exec
+    # This is called by controller.instance_exec, which changes self to the
+    # controller instance. That makes the instance variables unavailable to the
+    # proc via closure. Copying the instance variables to local variables before
+    # the proc is returned keeps them in scope for the returned proc.
     value = @value
     block = @block
     if value.respond_to?(:call)
--- a/lib/inertia_rails/renderer.rb
+++ b/lib/inertia_rails/renderer.rb
@@ -27,11 +27,11 @@ module InertiaRails
     private
 
     def props
-      _props = ::InertiaRails.shared_data(@controller).merge(@props).select do |key|
+      _props = ::InertiaRails.shared_data(@controller).merge(@props).select do |key, prop|
         if rendering_partial_component?
           key.in? partial_keys
         else
-          true
+          !prop.is_a?(InertiaRails::Lazy)
         end
       end
 
--- a/lib/inertia_rails/renderer.rb
+++ b/lib/inertia_rails/renderer.rb
@@ -27,13 +27,13 @@ module InertiaRails
     private
 
     def props
-      only = (@request.headers['X-Inertia-Partial-Data'] || '').split(',').compact.map(&:to_sym)
-
-      _props = ::InertiaRails.shared_data(@controller).merge(@props)
-
-      _props = (only.any? && @request.headers['X-Inertia-Partial-Component'] == component) ?
-        _props.select {|key| key.in? only} :
-        _props
+      _props = ::InertiaRails.shared_data(@controller).merge(@props).select do |key|
+        if rendering_partial_component?
+          key.in? partial_keys
+        else
+          true
+        end
+      end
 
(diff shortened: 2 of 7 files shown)


lib/inertia_rails/lazy.rb
```diff
@@ -0,0 +1,23 @@
+class InertiaRails::Lazy
+  def initialize(value = nil, &block)
+    @value = value
+    @block = block
+  end
+
+  def call
```
bknoles, line 7: call isn't actually used in the library, other than in `prop.respond_to?(:call)` in the renderer.

i decided to implement it anyway and add some explicit test coverage, since the rendering specs don't actually test it

lib/inertia_rails/lazy.rb
```diff
@@ -0,0 +1,23 @@
...
+    to_proc.call
+  end
+
+  def to_proc
```
bknoles, line 11: there might be a more elegant way to express this, but it does get the tests green!

lib/inertia_rails/renderer.rb
```diff
@@ -27,13 +27,13 @@ def render
...
-      _props = (only.any? && @request.headers['X-Inertia-Partial-Component'] == component) ?
-        _props.select {|key| key.in? only} :
-        _props
+      _props = ::InertiaRails.shared_data(@controller).merge(@props).select do |key, prop|
+        if rendering_partial_component?
+          key.in? partial_keys
+        else
+          !prop.is_a?(InertiaRails::Lazy)
+        end
+      end
```
bknoles, line 36: stuffing some of this code into methods should reveal the intent a bit better

spec/dummy/app/controllers/inertia_render_test_controller.rb
```diff
@@ -17,4 +17,15 @@ def view_data
...
+  def lazy_props
+    render inertia: 'TestComponent', props: {
+      name: 'Brian',
+      sport: InertiaRails.lazy('basketball'),
+      level: InertiaRails.lazy do
+        'worse than he believes'
+      end,
+      grit: InertiaRails.lazy(->{ 'intense' })
+    }
+  end
```
bknoles, line 30: This is the API for using lazy props... it works with a plain value, a block, or a block/lambda passed as a value

lib/inertia_rails/renderer.rb
BrandonShar: boy that was confusing. Good call.

lib/inertia_rails/lazy.rb
```diff
@@ -0,0 +1,23 @@
...
+    # copy to local variables so they are in scope for controller.instance_exec
```
BrandonShar, line 12: I don't understand the need for this.

lib/inertia_rails/renderer.rb
```diff
@@ -27,13 +27,13 @@ def render
...
-      _props = ::InertiaRails.shared_data(@controller).merge(@props)
-
-      _props = (only.any? && @request.headers['X-Inertia-Partial-Component'] == component) ?
-        _props.select {|key| key.in? only} :
-        _props
+      _props = ::InertiaRails.shared_data(@controller).merge(@props).select do |key, prop|
+        if rendering_partial_component?
+          key.in? partial_keys
+        else
+          !prop.is_a?(InertiaRails::Lazy)
```
BrandonShar, line 34: This sets off my duck-typing alarm bells a bit, but off the top of my head, I can't think of a better option since the alternative is various "primitives"...

BrandonShar (approved):
👍 

This was simpler than I expected. I kept looking for the part that made it all work before realizing that this was it haha.

lib/inertia_rails/lazy.rb
bknoles: i added a more detailed comment... things get weird because this ends up being called with `controller.instance_exec`. that changes the value of `self` while the proc gets executed, and the instance variables are unavailable within them.

copying the instance variables to local variables in the method keeps them available to the Proc via closure.

lib/inertia_rails/renderer.rb
bknoles: I could definitely see a future where _all_ props become some kind of polymorphic prop class so this logic gets pulled out of the renderer and there is a more consistent treatment of the "kinds" of props (plain data, blocks, and lazy)

spec/dummy/app/controllers/inertia_render_test_controller.rb
snaptopixel: @reinink @bknoles please add this to the docs, had to dig a bit to find the api after seeing it in the changelog https://inertiajs.com/partial-reloads

snaptopixel: Oops, just saw the PR to update the Inertia site, nevermind me ☺️

BrandonShar: Yep! Awaiting merge right here: https://github.com/inertiajs/inertiajs.com/pull/143

😄

PR #53, opened 2020-11-24
Declare InertiaRails module before declaring the Lazy class

bknoles:
Fixes an issue noticed by @buhrmi here:

https://github.com/inertiajs/inertia-rails/commit/93ba80b6e1a72e5664c82156f56fc407a01a461f


bknoles:
couldn't figure out how to reproduce the issue in the test suite, but it happens when you try to boot an app that includes the gem.

PR #54, opened 2020-11-28
Create inertia route helper

BrandonShar:
This PR creates the `inertia` route helper verb so that our routes file can contain an entry like this:

```ruby
inertia 'about' => 'AboutComponent'
```

Which will route the `/about` URI directly to your `Pages/AboutComponent` without setting up the boilerplate of a controller and render. By design it does not support passing props, if you need that, continue to use the normal controller setup.

**Background:**
@euggra and I were working on an inertia rails app that had some static content-only pages like an About page. We wanted to write these pages in react like the rest of the app, but had to write too much boilerplate for my liking just to render a prop-less component. Turns out that inertia-laravel already had a solution for this, so it was time for us to join them in the future!

### DIFF
--- a/lib/inertia_rails.rb
+++ b/lib/inertia_rails.rb
@@ -20,3 +20,12 @@ end
 module InertiaRails
   class Error < StandardError; end
 end
+
+class ActionDispatch::Routing::Mapper
+  def inertia(args, &block)
+    route = args.keys.first
+    component = args.values.first
+
+    get(route => 'inertia_rails/static#static', defaults: {component: component})
+  end
+end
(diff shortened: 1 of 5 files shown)


lib/inertia_rails.rb
```diff
@@ -20,3 +20,12 @@
 module InertiaRails
   class Error < StandardError; end
 end
+
+class ActionDispatch::Routing::Mapper
```
bknoles, line 24: is this the best practice method for adding methods to Rails classes?

this is a genuine question... i know there are several ways to extend classes in Ruby.

BrandonShar: Good call second-guessing this. It looks like doing it this way in a gem creates a risk that if things are executed in the wrong order, we could end up defining that class before rails. 

Luckily we already have a pattern for doing this in our gem that I missed so I changed this to mimic that.

bknoles (approved):
looks good!

Issue #55, opened 2020-12-19
Error when trying to run the specs


christoomey:
Hey inertia-rails team, first of all, thanks for the great work on this project 👋!

I unfortunately was unable to get the tests running locally as I'm hitting the following error:

<details>
  <summary><strong>Error output when running `bundle exec rake`</strong></summary>

```
An error occurred while loading rails_helper.
Failure/Error: 'Cache-Control' => "public, max-age=#{1.hour.to_i}"

NoMethodError:
  undefined method `hour' for 1:Integer
# ./spec/dummy/config/environments/test.rb:19:in `block in <top (required)>'
# ./spec/dummy/config/environments/test.rb:6:in `<top (required)>'
# ./spec/dummy/config/environment.rb:5:in `<top (required)>'
# ./spec/rails_helper.rb:4:in `require'
# ./spec/rails_helper.rb:4:in `<top (required)>'
No examples found.
No examples found.
```
</details>

I'm guessing there's some setup step needed in the spec/dummy app or similar, but not sure what it is. I'd be more than happy to document or add to `bin/setup` if you point me in the right direction.

ledermann:
It seems this is caused by Rails 6.1, which requires this line to be added on top of `spec/dummy/config/environments/test.rb`:

```ruby
require 'active_support/core_ext/integer/time'
```

See http://railsdiff.org/6.0.3.4/6.1.0

BrandonShar:
Thanks @christoomey and @ledermann! Great issue info and solution.

PR #56, opened 2020-12-19
Add redirect_back w/ inertia errors support

christoomey:
This builds on the work in 40f36a5580a3362ed2e9dbd23373b61e82dfc98d which added an inertia-aware wrapper around `redirect_to`.

This change implements the same inertia errors handling for
`redirect_back`, which I've found to be particularly useful when working
with Inertia, especially with error handling.

---

Note: I, unfortunately, wasn't able to get the specs running locally (see notes in https://github.com/inertiajs/inertia-rails/issues/55).

Also, I noticed that this behavior is currently [undocumented on the inertia site](https://inertiajs.com/forms#server-side-validation), and
I only discovered this feature by reading through recent PRs. I'd be happy to add to
the documentation, assuming that's something that would of interest.

### DIFF
--- a/lib/inertia_rails/controller.rb
+++ b/lib/inertia_rails/controller.rb
@@ -20,17 +20,31 @@ module InertiaRails
       end
     end
 
+    private
+
     def inertia_location(url)
       headers['X-Inertia-Location'] = url
       head :conflict
     end
 
     def redirect_to(options = {}, response_options = {})
-      if (inertia_errors = response_options.fetch(:inertia, {}).fetch(:errors, nil))
+      capture_inertia_errors(response_options)
+      super(options, response_options)
+    end
+
+    def redirect_back(fallback_location:, allow_other_host: true, **options)
+      capture_inertia_errors(options)
+      super(
+        fallback_location: fallback_location,
+        allow_other_host: allow_other_host,
+        **options,
+      )
+    end
+
+    def capture_inertia_errors(options)
+      if (inertia_errors = options.dig(:inertia, :errors))
         session[:inertia_errors] = inertia_errors
       end
-
-      super(options, response_options)
     end
   end
 end
--- a/spec/dummy/app/controllers/inertia_test_controller.rb
+++ b/spec/dummy/app/controllers/inertia_test_controller.rb
@@ -33,6 +33,13 @@ class InertiaTestController < ApplicationController
     redirect_to empty_test_path, inertia: { errors: { uh: 'oh' } }
   end
 
+  def redirect_back_with_inertia_errors
+    redirect_back(
+      fallback_location: empty_test_path,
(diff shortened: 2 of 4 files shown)


lib/inertia_rails/controller.rb
```diff
@@ -20,17 +20,31 @@ def inertia_share(**args, &block)
       end
     end
 
+    private
```
christoomey, line 23: Somewhat unrelated to the specific change, but I _think_ we'd always want these methods to be private. Happy to revert if there are any concerns with this.

spec/dummy/config/environments/test.rb
```diff
@@ -16,7 +16,7 @@
   # Configure public file server for tests with Cache-Control for performance.
   config.public_file_server.enabled = true
   config.public_file_server.headers = {
-    'Cache-Control' => "public, max-age=#{1.hour.to_i}"
+    'Cache-Control' => "public, max-age=#{3600}"
```
christoomey, line 19: This change was enough to get the specs running locally for me. Happy to revert, but wanted to include in case this feels like a worthwhile change.

lib/inertia_rails/controller.rb
BrandonShar: Wow, TIL that I did not understand method visibility in ruby as well as I thought I did; very different than PHP. Cool stuff!

BrandonShar (approved):
Thanks for the PR @christoomey (and good to hear from you again).

This makes good sense to me! :+1:

Unless you're in a hurry/available to battle test the release, I'll probably wait until Monday to actually tag this.

christoomey:
No hurry on my end (and nice to connect as well 👋). I implemented this directly in my own codebase, but felt like it was worth upstreaming, so wanted to share here.

BrandonShar:
For sure. And for your other question, yes rails docs contributions are always welcome on https://github.com/inertiajs/inertiajs.com . We have a bad habit of occasionally batch adding things instead of as they release 😄 .

christoomey:
Hey @BrandonShar, happy new year 👋 Wanted to check back in and see if this was still in consideration.

BrandonShar:
gah, sorry; totally slipped my mind!

BrandonShar:
@christoomey just tagged `1.9.0`, let me know if anything seems amiss.

And thanks for the bump!

christoomey:
Awesome, thanks!

lib/inertia_rails/controller.rb
ledermann: I'm a bit late here, but `redirect_to` needs to be `public`, which is the Rails default. Changing it to `private` causes issues when using Devise, so I reverted it in #60.

christoomey: Sorry for the issue here!

PR #60, opened 2021-02-10
Fix compatibility with "responders" gem (and thus "devise")

ledermann:
In v1.9.0, the methods `redirect_to`(and `redirect_back`) were changed to be `private` by #56. IMHO this is not correct, because these methods are `public` by default:

```ruby
irb(main):001:0> ApplicationController.new.public_methods.include?(:redirect_to)
=> true
irb(main):002:0> ApplicationController.new.public_methods.include?(:redirect_back)
=> true
```

The changed visibility causes issues when using the gem [devise](https://github.com/heartcombo/devise), which uses the gem [responders](https://github.com/heartcombo/responders/) internally - see [this error](https://github.com/ledermann/pingcrm/pull/487/checks?check_run_id=1738638934) in my PingCRM demo application:

```ruby
LoginTest#test_Login_with_valid_credentials_will_be_successful:
NoMethodError: private method `redirect_to' called for #<Users::SessionsController:0x00000000009dd0>
    app/controllers/users/sessions_controller.rb:9:in `create'
```

This PR restores the original method visibility and adds testing against the `responders` gem.

### DIFF
--- a/inertia_rails.gemspec
+++ b/inertia_rails.gemspec
@@ -33,4 +33,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "rails-controller-testing"
   spec.add_development_dependency "sqlite3"
   spec.add_development_dependency "appraisal"
+  spec.add_development_dependency "responders"
 end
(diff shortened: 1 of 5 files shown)


bknoles:
thanks @ledermann ! i'll release a new version

PR #62, opened 2021-02-11
Better setup specs warning

### LINKED ISSUE #61: Misleading "Inertia test helpers aren't set up!" warning
I've been using the spec helper and I've run into an issue where I'm seeing "Inertia test helpers aren't set up!", despite the fact that they are indeed set up. In my case I ran into this when a `before_action` that I was not expecting to run did end up running, and thus I never got to the `render(inertia ...` call.

This seems to be due to the fact that [the stubbing of `render` method](https://github.com/inertiajs/inertia-rails/blob/5c71d77dec869e130356ec4d259c94f4adbb30b2/lib/inertia_rails/rspec.rb#L63-L65) is done lazily when the `render(inertia ...` is actually called.

Presuming these spec helpers are intended for use in applications and not just for the library itself (which I really hope is the intention as they're great!), I think it would be great to look into fixing this as it can lead to a confusing experience. I'd be happy to poke around and see if I could resolve this and open a PR if I could, but wanted to start with an issue to see if there are any existing thoughts or constraints I'm unaware of.

Thanks again for the great library!

bknoles:
Resolves #61 

As noted by @christoomey , the error message for ensuring users set `inertia: true` in RSpec specs can be misleading.

This PR changes the code to check the actual example metadata before warning about the test setup.

Additionally, it adds a warning when you create a spec that expects inertia but never creates an inertia renderer. This isn't directly part of #61, but is something that has tripped me up in the past.

### DIFF
--- a/lib/inertia_rails/rspec.rb
+++ b/lib/inertia_rails/rspec.rb
@@ -49,7 +49,7 @@ module InertiaRails
       protected 
       
       def inertia_tests_setup?
-        @_inertia_render_wrapper.present?
+        ::RSpec.current_example.metadata.fetch(:inertia, false)
       end
     end
   end
--- a/lib/inertia_rails/rspec.rb
+++ b/lib/inertia_rails/rspec.rb
@@ -35,6 +35,9 @@ module InertiaRails
       def inertia
         raise 'Inertia test helpers aren\'t set up! Make sure you add inertia: true to describe blocks using inertia tests.' unless inertia_tests_setup?
 
+        if @_inertia_render_wrapper.nil? && !::RSpec.configuration.inertia[:skip_missing_renderer_warnings]
+          warn 'WARNING: the test never created an Inertia renderer. Maybe the code wasn\'t able to reach a `render inertia:` call? If this was intended, or you don\'t want to see this message, set ::RSpec.configuration.inertia[:skip_missing_renderer_warnings] = true'
+        end
         @_inertia_render_wrapper
       end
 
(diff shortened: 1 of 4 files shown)


lib/inertia_rails/rspec.rb
```diff
@@ -49,14 +52,17 @@ def inertia_wrap_render(render)
       protected 
       
       def inertia_tests_setup?
-        @_inertia_render_wrapper.present?
+        ::RSpec.current_example.metadata.fetch(:inertia, false)
```
bknoles, line 55: There might be a better RSpec way to get at this data, but this seems to work ok

lib/inertia_rails/rspec.rb
```diff
@@ -49,14 +52,17 @@ def inertia_wrap_render(render)
...
       end
     end
   end
 end
 
 RSpec.configure do |config|
   config.include ::InertiaRails::RSpec::Helpers
+  config.add_setting :inertia, default: {
```
bknoles, line 63: created this as an inertia scoped hash in case it needs to grow in the future.

My quick research led me to `config.add_setting`. If there's a more RSpec-y way to do this, happy to use it.

christoomey:
Looks good to me! Thanks for the quick update 🚀

spec/inertia/rspec_helper_spec.rb
```diff
@@ -54,4 +70,48 @@
...
+        expect(fake_std_err.messages[0].chomp).to(eq(warn_message))
+      ensure
+        $std_err = original_stderr
+      end
+    end
+
+    context 'with the :skip_missing_renderer_warnings setting set to true' do
+      before {
+        @original = ::RSpec.configuration.inertia[:skip_missing_renderer_warnings]
+        ::RSpec.configuration.inertia[:skip_missing_renderer_warnings] = true
```
BrandonShar, line 99: This feels a little cumbersome, but I imagine will be pretty uncommon. If this is rspec's way to handle configs though, works for me!

BrandonShar (approved):
Awesome stuff! I was so happy with error message originally, great to make sure it won't trick people any longer :smile:

PR #64, opened 2021-03-19
Set up generator

bknoles:
Adds a generator for installing a few necessary files for a React based inertia_rails project:

* An `inertia.jsx` pack file with CSRF set up that renders an Inertia app, and adds it into `application.html.erb`
* A controller, route, and React component for a "hello world!" inertia-fied page
* Adds the appropriate inertiajs packages via Yarn
* Stubbed, non-functional rake tasks for Vue and Svelte installs (we use React internally so we haven't built out these installers yet)

The task is run via:

```
./bin/rails inertia_rails:install:react
./bin/rails inertia_rails:install:vue # doesn't do anything yet
./bin/rails inertia_rails:install:svelte # doesn't do anything yet
```

### DIFF
--- a/lib/generators/inertia_rails/install/controller.rb
+++ b/lib/generators/inertia_rails/install/controller.rb
@@ -1,0 +1,7 @@
+class InertiaExampleController < ApplicationController
+  def index
+    render inertia: 'InertiaExample', props: {
+      name: 'World',
+    }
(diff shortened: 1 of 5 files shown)


BrandonShar (approved):
Crazy how little code this ended up being!

PR #65, opened 2021-03-23
feat: Implement generator for Vue apps

ElMassimo:
### Description 📖 

This pull request is a follow-up to #64, and implements the `inertia_rails:vue` installer.

### Bug Fix 🐞 

The installer was __always bailing out__, as `installable?` was returning `nil` when both checks passed.

### Notes ✏️ 

Moved the React specific files to a folder to keep things organized, and provide some consistency in the file names across installers.

### Future Work 🔮 

Would love it if these installers would also support [Vite Ruby](https://github.com/ElMassimo/vite_ruby).

For a full example, check this [Ping CRM](https://github.com/ElMassimo/pingcrm-vite) version with Vite.js


lib/generators/inertia_rails/install_generator.rb
```diff
@@ -32,6 +33,8 @@ def installable?
 
         return false
       end
+
+      true
```
BrandonShar, line 37: That's what we get for refactoring without tests 🙄 

Thanks for catching this!

BrandonShar (approved):
:+1: Looks great! Thanks @ElMassimo

bknoles:
Mil gracias @ElMassimo 🙏  ! 

We had hope to get PR's for Vue and Svelte, but didn't expect to get one so quickly! i'll cut a release today.

bknoles:
Released as `1.11.0`

https://rubygems.org/gems/inertia_rails/versions/1.11.0

PR #69, opened 2021-06-26
Implement generator for Svelte apps

ukazap:


bknoles (approved):
look great! i've never personally used Svelte, but i pulled this down and fired up an app and everything worked!

thank you for contributing this back to the project @ukazap !

ukazap:
Thanks for merging the changes @bknoles, this gem will be a boost for my future projects as well!

Issue #58, opened 2021-01-12
Intermittent 302 on DELETE requests


christoomey:
I recently ran into an issue where the "Sign Out" button of my app would occasionally result in an inertia modal w/ 404 error page.

I opened this as https://github.com/inertiajs/inertia/issues/419 on the main Inertia repo, but they kindly helped me dig a bit deeper and I've now confirmed that the issue is that response status code.

At this point I'm guessing the issue is somewhere in my app / middleware stack, but I wanted to share here in case you folks have any thoughts as to how this might be happening.

---

I'm specifically running into this with the "Sign Out" button in my app. The problem flow is.

- User clicks "Sign Out" button
- We fire an Inertia DELETE to /sign_out
- The delete is processed, the user is signed out, and the response comes back as a 302 w/ location /sign_in
- Instead of issuing a GET to /sign_in, a DELETE request is issued. This ends up not matching any route, so a 404 error is displayed in the error modal

I've found this tricky to reproduce as most cases send the correct 303 status code, but I have been able to reproduce it.

For reference, here's my middleware stack.

<details>
  <summary>Rack middleware stack</summary>

```
use Rack::Cors
use Honeybadger::Rack::UserInformer
use Honeybadger::Rack::UserFeedback
use Honeybadger::Rack::ErrorNotifier
use Rack::Rewrite
use ActionDispatch::SSL
use Rack::Sendfile
use ActionDispatch::Static
use Rack::Deflater
use ActionDispatch::Executor
use ActiveSupport::Cache::Strategy::LocalCache::Middleware
use RailsAutoscaleAgent::Middleware
use Rack::Runtime
use Rack::MethodOverride
use ActionDispatch::RequestId
use Rack::Timeout
use RequestStore::Middleware
use ActionDispatch::RemoteIp
use Rails::Rack::Logger
use ActionDispatch::ShowExceptions
use ActionDispatch::DebugExceptions
use Rollbar::Middleware::Rails::RollbarMiddleware
use ActionDispatch::Callbacks
use ActionDispatch::Cookies
use ActionDispatch::Session::CookieStore
use ActionDispatch::Flash
use ActionDispatch::ContentSecurityPolicy::Middleware
use Rack::Head
use Rack::ConditionalGet
use Rack::ETag
use Rack::TempfileReaper
use Clearance::RackSession
use Warden::Manager
use Coverband::BackgroundMiddleware
use Rack::Attack
use ScoutApm::Middleware
[...]

PR #70, opened 2021-06-26
Make middleware thread safe

caifara:
Issue #58 happens when a delete request (or similar) and a get request are processed in different threads concurrently.

In the middleware @env isn't thread safe as it is an instance variable of the application (not scoped within the thread). That implicates that a new get request following a delete request on another thread would override @env during the delete request cycle thus not returning 303 but the Rails default 302.

I have two commits in this pr. The first one solves the problem by simply handing the env variable to every method depending on it. The second one uses a new class to be able to use the @env abstraction. In my opinion that approach is way cleaner and won't encourage other developers to reintroduce instance variables in the middleware itself.

### DIFF
--- a/spec/inertia/middleware_spec.rb
+++ b/spec/inertia/middleware_spec.rb
@@ -28,6 +28,24 @@ RSpec.describe InertiaRails::Middleware, type: :request do
 
       it { is_expected.to eq 303 }
     end
+
+    it 'is thread safe' do
+      delete_request_proc = -> { delete redirect_test_path, headers: { 'X-Inertia' => true } }
+      get_request_proc = -> { get empty_test_path }
+
+      statusses = []
+
+      threads = []
+
+      100.times do
+        threads << Thread.new { statusses << delete_request_proc.call }
+        threads << Thread.new { get_request_proc.call }
+      end
+
+      threads.each(&:join)
+
+      expect(statusses.uniq).to eq([303])
+    end
   end
 
   context 'a request not originating from inertia' do
--- a/lib/inertia_rails/middleware.rb
+++ b/lib/inertia_rails/middleware.rb
@@ -5,70 +5,69 @@ module InertiaRails
     end
 
     def call(env)
-      @env = env
-
       status, headers, body = @app.call(env)
       request = ActionDispatch::Request.new(env)
 
       ::InertiaRails.reset!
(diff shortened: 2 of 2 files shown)


bknoles:
Thank you @caifara ! You are correct on the problem, and I really like the `IntertiaRailsRequest` refactor. Even just creating the `response` method makes the code cleaner IMO.

I spent some time looking into middleware thread safety after seeing this, and for anyone reading this later, a bit more detail on the root issue is (by my understanding):

* Middlewares are instantiated once per process
* Threads are created within a process and share the same already-instantiated middleware classes
* Therefore, any instance variables created on a middleware class will be shared across all threads within the process
* InertiaRails was setting an `@env` instance variable on the middleware class _on every request_. This led to a single `@env` being used by every concurrently running request within a process.
* Generally speaking, mutating instance variables on a middleware class within the `call(env)` method makes the middleware not thread safe. (There are workarounds, but it's a good rule of thumb)

Resources I found helpful:
https://stackoverflow.com/questions/23028226/rack-middleware-and-thread-safety
https://crypt.codemancers.com/posts/2018-06-07-frozen-middleware-with-rack-freeze/
https://tenderlovemaking.com/2012/06/18/removing-config-threadsafe.html

spec/inertia/middleware_spec.rb
```diff
@@ -28,6 +28,24 @@
...
+      threads = []
+
+      100.times do
+        threads << Thread.new { statusses << delete_request_proc.call }
+        threads << Thread.new { get_request_proc.call }
+      end
+
+      threads.each(&:join)
+
+      expect(statusses.uniq).to eq([303])
```
bknoles, line 47: I ran this test with the original buggy code, and it will occasionally pass by random luck. I'm not worried about that, since the code structurally prevents the issue really well and I would not expect to see a regression here, but figure it's worth pointing out in case it ever pops up as confusing in the future

bknoles (approved):

bknoles:
Released in a v1.11.1!

https://rubygems.org/gems/inertia_rails/versions/1.11.1

Thanks again @caifara !

caifara:
@bknoles We're really fond of inertia and inertia-rails, glad to be able to do something in return! Thanks for releasing!

Issue #71, opened 2021-07-26
Making CSRF protection work out of the box


buhrmi:
At the moment, [manual steps](https://discord.com/channels/592327939920494592/759114256817455145/773925412626628618) are required to make CSRF protection work with Rails and Inertia.

To make CSRF protection work out of the box I'm proposing to add the following to the inertia-rails adapter:
1. Make every non-inertia request also send the current csrf token in a `XSRF-TOKEN` cookie, that will automatically be picked up by Axios.
2. Overwrite the [`request_authenticity_tokens`](https://github.com/rails/rails/blob/83217025a171593547d1268651b446d3533e2019/actionpack/lib/action_controller/metal/request_forgery_protection.rb#L302) method to also check the `X-XSRF-TOKEN` header value (in addition to the `X-CSRF-TOKEN` header), since that is what Axios sends by default.

BrandonShar:
Per this comment on the linked PR (https://github.com/inertiajs/inertia-rails/pull/72#issuecomment-941298861) I believe we decided that this wasn't necessary and is somewhat out of scope for Inertia.

As such, I'm going to close this issue, but please reopen if I've misunderstood!

bknoles:
I'm going to reopen this issue. Our workaround, which was to add the CSRF token to an `X-CSRF-Token` header in the axios defaults ([see here](https://github.com/inertiajs/inertia-rails/blob/master/lib/generators/inertia_rails/install/react/inertia.jsx#L12)), fails to work on certain projects. I finally figured out why.

The workaround requires that NPM/yarn imports the axios package from inertia's dependencies. If axios is installed somewhere else in the dependency tree, and NPM/yarn imports that instance instead of the one that comes with Inertia, then Inertia requests will never set the `X-CSRF-Token` header. I realized this because CSRF stopped working after I installed axios as an explicit dependency (because ESLint was yelling at me about it).

The solution I added to the app was almost the same as @buhrmi 's original one. I added the `XSRF-Token` cookie in a controller hook. Axios will grab this cookie and send it with all requests. But instead of overwriting the `request_authenticity_tokens` method, I added middleware that copies an `X-XSRF-Token` header (which Axios sends) into an `X-CSRF-Token` header (which Rails looks for). I like this better than monkey patching, but I'm open to hear any downsides I haven't thought of.

[...]

PR #72, opened 2021-08-08
make CSRF protection work out of the box (#71)

buhrmi:
This PR removes the need for the [manual steps](https://discord.com/channels/592327939920494592/759114256817455145/773925412626628618) to make CSRF protection work. It does this by storing the current form_authenticity_token in the `XSRF-TOKEN` cookie (which is picked up by Axios automatically) and adding the the `X-XSRF-TOKEN` header value (which is sent by Axios by default) in the list of possible authenticity tokens.

### DIFF
--- a/lib/inertia_rails.rb
+++ b/lib/inertia_rails.rb
@@ -1,6 +1,7 @@
 require 'inertia_rails/renderer'
 require 'inertia_rails/engine'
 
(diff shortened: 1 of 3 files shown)


louis-smit:
Hey everyone. Is there any update on this? Seems to work great and would be cool to have it merged in🙂

PR #73, opened 2021-10-07
WIP: SSR integration

zealot128:
Just want to let you know, I am working on a SSR integration.

How it works:  
It is based on the SSR guide recently released to Inertia backer's and Vite's SSR guide amalgamated into this prototype/ProofOfConcept. When rendering a template for the first time, it will check, if a SSR server is running and try to ``post /render`` to it with the ``page`` as a json payload. The SSR server is expected to respond with a JSON object of ``{ head: [], body: "" }``. The plugin then puts the ``body`` as the complete template, and supplies the ``head`` as instance variables for the application layout. If any error occurs alongside, the plugin will fallback to the default handling of rendering inertia template with div#app. 

### Configuration:

SSR is opt-in: 

```ruby
# config/initializers/inertia.rb
InertiaRails.ssr_enabled = true
InertiaRails.ssr_host = 'localhost'
InertiaRails.ssr_port = 23526
```

To add headers from rendered files (supplied by inertia-head), also add them to your application layout:

```erb
    <% if @ssr_headers %>
      <!-- SSR-headers -->
      <% @ssr_headers.each do |head| %>
        <%= raw head %>
      <% end %>
    <% end %>
 </head>
```

- Start the SSR server as well as the Rails Server., e.g. ``node server.js``.
- If SSR is not required/wished for a specific action, you can pass  ``render inertia: {..., ssr: false }``

## Development:
Missing:
- [ ] Tests.. hadn't even looked into the repo (happy if you give me any pointers which tests to add)
- [ ] Docs
- [ ] SSR script/guide/docs for the webpacker folks (maybe other people can join in here)

(How to make a SSR also depends a lot on your bundler, Webpack, Vite etc., we are using Vite at the moment)

<details>
<summary>Our Vite + Inertia + Vue SSR server:</summary>

```js
// server.js
const fs = require("fs")
const path = require("path")
const express = require("express")
const { createServer: createViteServer } = require("vite")

async function createServer() {
  const app = express()
  app.use(express.json())

  const vite = await createViteServer({
    server: {
      middlewareMode: "ssr",
      port: 23526,
      host: '0.0.0.0',
      hmr: false,
      // hmr: { port: 23526, protocol: "wss", host: "jobsdaheim.swi.pludoni.com" },
      https: false,
    },
  })
  app.use(vite.middlewares)

  app.post("/render", async (request, response, next) => {
    try {
      const render = await vite.ssrLoadModule("~/ssr")
      const out = await render.default(request.body)
      response.json(out)
[...]

PR #74, opened 2022-02-12
Inertia SSR

BrandonShar:
Long overdue, we're finally readying for Inertia Rails SSR!

First off, a huge thanks to  @zealot128 for their work on #73 . I had already started this version so I wanted to push forward with it, but their code was a great help in ensuring we were all thinking about things in a similar way.

Part of the hold up here ha been that Rail's Webpacker is not quite as fluid as Laravel Mix so I wanted to wait until I felt like we had a solid way of recommending the JS setup. After a lot of tinkering, I'm happy to say that I think we do!

First off, the InertiaRails portion of this is very simple, InertiaRails gets 2 new config values that can be set (the values shown are the defaults):

### Backend

```ruby
   config.ssr_enabled = false
   config.ssr_url = 'http://localhost:13714'
end
```

### Frontend

Now comes the harder part, the javascript:

First things first, let's add a new dependency.
```
yarn add webpack-node-externals
```

More information can be found in the Inertia SSR docs: https://inertiajs.com/server-side-rendering

Unfortunately, Laravel Mix doesn't exist for rails, so the docs can't help us much further (though they do give us the contents of one more file).

I created/modified the following files:

**Create** `app/javascript/ssr/ssr.jsx`
```js
import React from 'react'
import ReactDOMServer from 'react-dom/server'
import { createInertiaApp } from '@inertiajs/inertia-react'
import createServer from '@inertiajs/server'

createServer((page) => createInertiaApp({
  page,
  render: ReactDOMServer.renderToString,
  resolve: name => require(`../Pages/${name}`),
  setup: ({ App, props }) => <App {...props} />,
}))
```

This comes directly from the Inertia docs and serves as the entry point for Inertia SSR.

**Create** `config/webpack/ssr.js`
```js
process.env.NODE_ENV = 'ssr'

const { environment } = require('@rails/webpacker')
const nodeExternals = require('webpack-node-externals')

const config = environment.toWebpackConfig();

module.exports = {
  ...config,
  output: {
    ...config['output'],
    filename: 'ssr.js',
  },
  target: 'node',
  externals: [nodeExternals()],
}

```

This file is unique to rails but is similar in spirit to Inertia's `ssr.mix.js` file. The basic idea is that the SSR node process is distinct from our usual JS, so we want to configure it a bit differently.

**Edit** `config/webpacker.yml` and add the following entry:
```yml
ssr:
  <<: *default
  compile: false
  extract_css: true

  source_path: app/javascript
  source_entry_path: ssr
  public_root_path: public
  public_output_path: ssr
```

This entry will allow us to keep our ssr js separate since it's not really a pack in the rails sense.

**Edit** `babel.config.js`
Change the individual lines that look like
```js
var validEnv = ['development', 'test', 'production']
var isProductionEnv = api.env('production')
```
to 
```js
var validEnv = ['development', 'test', 'production', 'ssr']
var isProductionEnv = api.env('production') || api.env('ssr')
```
This let's us specify our new ssr node env, but otherwise treat it like production.

**Edit** `app/views/layouts/application.html.erb` and add the following line inside of the head tags
```erb
<%= inertia_headers %>
```
This allows Inertia SSR to add to the document head

And finally

**Edit** `package.json` and add (or modify if you already have one) to add a scripts section
```json
"scripts": {
    "ssr": "NODE_ENV=ssr RAILS_ENV=ssr ./bin/webpack && node ./public/ssr/ssr.js"
  }
```
This will allow you to run the SSR server locally with the command `yarn run ssr` (but keep in mind that it does NOT hot reload, you will need to kill and re-run this command when you make JS changes!)

And that should be that! Now kickback and disable your javascript!

If these changes work well for most users, we'll get them moved into the templates.

### Outstanding questions

For those of you who need SSR, do you prefer to use SSR in development? As of right now, the best workflow from an ease of use standpoint would be to use traditional Inertia in development and SSR only in production. Obviously, this runs of the risk of accidentally using client only javascript or otherwise breaking the production build despite working fine in development. For those of you using this, is that an acceptable tradeoff or should we invest some time in making SSR in development a better experience?


bknoles:
I'm curious to know how Inertia based SSR does with React hydration issues. We've seen this in a couple places:

* If the server rendered HTML doesn't match the client side "hydrated" HTML. I've seen this when pre-rendering pages at "dynamic" URLS. An example would be using react-snap with a React router based app that fetches a list of "posts" from an API and has a dynamically created route for each post. Since Rails actually handles all the routing, I would not expect this to be an issue for Inertia SSR.
* If the React app creates HTML outside of the React app root, then we get "orphaned" DOM nodes. We've seen this plague apps with portal based modals. I would assume an Inertia app with SSR _would) be vulnerable to this type of hydration issue.

If I can carve out some time, I'd like to test those on a dummy app. In the meantime, any theory based thoughts on how Inertia will fare here?

PR #75, opened 2022-04-28
With SSR enabled, X-Inertia Request should work as usual

99monkey:


BrandonShar:
Hey @99monkey thanks for the PR! Could you elaborate on the change a bit?

If I'm understanding correctly, this will allow Inertia to take over after the initial pageload and with the current setup with SSR enabled every page is a full reload?

bknoles:
This makes sense to me as default behavior. The two SSR use cases I can think of are:

1. Full page/meta previews for non-JS enabled scrapers (like a link preview on Facebook, for example). SSR on the "first page load only" satisfies that
2. Supporting users with JS disabled. This change doesn't affect them because the `X-Inertia` header will never get set by their browser

I also imagine that things like `preserveState` and `preserveScroll` don't work properly if SSR is always sending a full HTML page.

It also looks like this would match the behavior in the Laravel library:
https://github.com/inertiajs/inertia-laravel/blob/master/src/Response.php#L106

BrandonShar:
Makes sense to me! I'll add some test coverage and get it merged in asap.

PR #77, opened 2022-05-21
Fix: Use action_controller_base hook in initializer

0xDing:
Using inertia-rails in a Rails project where an `ActionController::API` controllers exists will trigger `undefined method 'helper_method' for ActionController::API:Class` error.

`action_controller` is called if the controller is an instance of `ActionController::Base` or `ActionController::API`.  And `ActionController::API` don't have `helper_method` method.

ref: rails/rails#28402

### DIFF
--- a/lib/inertia_rails/engine.rb
+++ b/lib/inertia_rails/engine.rb
@@ -8,7 +8,7 @@ module InertiaRails
     end
 
     initializer "inertia_rails.action_controller" do
-      ActiveSupport.on_load(:action_controller) do
+      ActiveSupport.on_load(:action_controller_base) do
         include ::InertiaRails::Controller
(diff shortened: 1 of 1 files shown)


BrandonShar:
Thanks @0xDing ! I'm going to add Rails 7 to our test checks and get this merged in!

BrandonShar (approved):

BrandonShar:
@bknoles this change looks good to me, but it will require us to drop Rails 5.0 support (the`action_controller_base` hook was only backported as far as Rails 5.1). Dropping 5.0 support seems fine to me since it's no longer supported by Rails either, but I figured it was worth a call-out and I'd say a major version bump. Agree?

PR #81, opened 2022-06-18
Mimic Rails behavior for instance props and rendering

BrandonShar:
Thanks to an excellent suggestion by @buhrmi, this PR would bring Rails view rendering behavior to Inertia Rails. It includes the following three new features:

## Instance Variables as Props
All instance variables defined during your controller method will be available to your inertia components as props.

**New Behavior:**
```ruby
def index
   @users = User.all
   
   render inertia: 'MyComponent'
end
```
**Is the same as currently writing:**
```ruby
def index
   render inertia: 'MyComponent', props: {
      users: User.all,
   }
end
```

## Automatically Determine Component Name
If you don't provide a component name, inertia will use Rails template logic to determine what it should be

**New Behavior:**
```ruby
   def index
      render inertia: nil # this syntax looks a little silly, so a new helper was added as shown below
      # OR
      render_inertia # this does the same thing as the render above, but looks a bit cleaner
   end
end
```
**is the same as currently writing:**
```ruby
   def index
      render inertia: 'users/index'
   end
end
```

## Make Inertia the Default Rendering Option
If your app is nearly all inertia-based, you can have the inertia render process be the default instead of rails views. Since this behavior is a bit magical feeling it's opt-in, not the default.

**New Behavior:**
```ruby
   config.default_render = true
end

   def index
      @users = User.all
   end
end
``` 
**is the same as currently writing:**
```ruby
   def index
      render inertia: 'users/index', props: {
         users: User.all,
      }
   end
end
```

### DIFF
--- a/lib/inertia_rails/controller.rb
+++ b/lib/inertia_rails/controller.rb
@@ -22,6 +22,18 @@ module InertiaRails
       end
     end
 
+    def render_inertia(component = nil)
+      render(inertia: component)
+    end
+
+    def default_render
+      if InertiaRails.default_render?
+        render_inertia
+      else
+        super
+      end
+    end
+
     def redirect_to(options = {}, response_options = {})
       capture_inertia_errors(response_options)
       super(options, response_options)
--- a/lib/inertia_rails/inertia_rails.rb
+++ b/lib/inertia_rails/inertia_rails.rb
@@ -32,6 +32,10 @@ module InertiaRails
     Configuration.ssr_url
   end
 
+  def self.default_render?
+    Configuration.default_render
+  end 
+
   def self.html_headers
     self.threadsafe_html_headers || []
   end
@@ -66,6 +70,7 @@ module InertiaRails
     mattr_accessor(:version) { nil }
     mattr_accessor(:ssr_enabled) { false }
     mattr_accessor(:ssr_url) { 'http://localhost:13714' }
+    mattr_accessor(:default_render) { false }
 
     def self.evaluated_version
       self.version.respond_to?(:call) ? self.version.call : self.version
--- a/lib/inertia_rails/renderer.rb
+++ b/lib/inertia_rails/renderer.rb
@@ -6,8 +6,8 @@ module InertiaRails
   class Renderer
     attr_reader :component, :view_data
 
-    def initialize(component, controller, request, response, render_method, props:, view_data:)
-      @component = component
+    def initialize(component = nil, controller, request, response, render_method, props:, view_data:)
+      @component = component || "#{controller.controller_path}/#{controller.action_name}"
       @controller = controller
       @request = request
       @response = response
@@ -38,7 +38,7 @@ module InertiaRails
     end
 
     def props
-      _props = ::InertiaRails.shared_data(@controller).merge(@props).select do |key, prop|
+      _props = ::InertiaRails.shared_data(@controller).merge(@controller.view_assigns).merge(@props).select do |key, prop|
         if rendering_partial_component?
           key.in? partial_keys
         else
--- a/lib/inertia_rails/renderer.rb
+++ b/lib/inertia_rails/renderer.rb
@@ -38,7 +38,7 @@ module InertiaRails
     end
 
     def props
-      _props = ::InertiaRails.shared_data(@controller).merge(@controller.view_assigns).merge(@props).select do |key, prop|
+      _props = ::InertiaRails.shared_data(@controller).merge(controller_instance_variables).merge(@props).select do |key, prop|
         if rendering_partial_component?
           key.in? partial_keys
         else
@@ -56,6 +56,10 @@ module InertiaRails
         url: @request.original_fullpath,
         version: ::InertiaRails.version,
       }
+    end
+
+    def controller_instance_variables
+      @controller.view_assigns.except('rendered_format', 'marked_for_same_origin_verification')
     end
 
     def deep_transform_values(hash, proc)
--- a/spec/dummy/app/controllers/inertia_rails_mimic_controller.rb
+++ b/spec/dummy/app/controllers/inertia_rails_mimic_controller.rb
@@ -1,0 +1,28 @@
+class InertiaRailsMimicController < ApplicationController
+  before_action :enable_inertia_default, only: :default_render_test
+
+  def instance_props_test
+    @name = 'Brandon'
+    @sport = 'hockey'
+
+    render inertia: 'TestComponent'
+  end
+
+  def default_render_test
+    @name = 'Brian'
+  end
+
+  def default_component_test
+    render inertia: nil
+  end
+
+  def default_component_shortcut_test
+    render_inertia
+  end
+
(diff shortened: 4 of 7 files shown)


BrandonShar:
Two questions for any reviewers:

1. **Do you think automatically including instance variables as props should also be opt-in as opposed to being the new default?**
2. **Should this be a major version bump?**

ledermann:
Great additions, I would like to add some comments from my perspective after trying this out in my [PingCRM](https://github.com/ledermann/pingcrm/) application:

### Instances variables as props

The current implementation always adds instance variables to the props, even if there are defined some explicitly. I don't think this is right because we lose control about what is being exposed. Maybe there are external gems used in the controller which define instance variables we would rather not be exposed. Example: In PingCRM, the gem "CanCanCan" sets the instance variables `@current_user` and `@current_ability`. Exposing them may be a security risk.

This said, I suggest exposing instance variables as props **only** if there is no `props` key given.
Furthermore, to prevent sensitive data from being inadvertently disclosed, I propose **not** setting this as default. Opt-in should be better, IMHO.

### Automatically Determine Component Name

Why not use `render inertia: true` instead of  `render inertia: nil`?  This doesn't look silly, so `render_inertia` could be omitted.

bknoles:
This is very cool. I really like how automatic component names provide a gentle suggestion for how the frontend code should be organized.

I agree with all of @ledermann 's suggestions, especially around the instance variable security risk. I think it should be opt-in at the very least, and there's a reasonable argument that it's too sharp of a knife. In a traditional Rails app, even though instance variables are automatically available to the view, the developer still has to explicitly choose to render them in order to expose them.

Building off the suggestions above, what about an API like:

```ruby
  config.render_instance_variables = true # truthiness enables the feature
  # Or
  config.render_instance_variables = {
    exclude: [:current_user, :current_ability]
  }
end

# And in the controller
def show
  @users = User.all
  # Doesn't render @users
  render inertia: true, props: { my: 'prop' }
  
  # renders @users, does not a render @current_admin prop, assumed to be set somewhere else in the app
  render inertia: true, except: [:current_admin]
end
```

That creates a fair amount of new permutations of the renderer code, so I could understand if that feels like too much complexity for too little API gain.

BrandonShar:
@ledermann I definitely agree with not mixing explicit props and implicit props, that's a good change.

 @bknoles with your change, as you mentioned, I'm a bit worried about having too many different apis. I think your example with abilities is a real concern though. Maybe we could have an `inertia_block` controller level method that functions the opposite of `inertia_share`? For some reason I feel like the config file is too far apart mentally from the controllers to be in charge of that logic. Or maybe we could provide some kind of pre-render hook that has a chance to remove props?

`render inertia: true` is also a bit nicer looking; I went for nil because it made the backend code easier to understand, but that might be a nicer api and allow us to drop the helper.

BrandonShar:
I really like that idea. Maybe we can use @ledermann's PingCRM app to test if that's feasible.

bknoles:
That's a very interesting idea @buhrmi . We can use `append_before_action`, but is there any way to be absolutely sure it's the last one in the chain? What if `CanCanCan` sets @current_ability in an `append_before_action` as well?

Some cursory research indicates that Bundler loads gems in the order in which they are defined in the Gemfile. That would mean we could never be certain our `append_before_action` runs last.

Maybe we could get around this by requiring the developer to set up the `before_action`? Something like:

```ruby
# In the application
class MyController < ApplicationController
  append_before_action :inertiafied_instance_variables # the name needs some work haha

  def show
    @users = User.all
    # renders @users as a prop, along with any shared data defined in ApplicationController
    render inertia: true
  end
end

# In the gem
module InertiaRails
  module Controller
    extend ActiveSupport::Concern

     def inertiafied_instance_variables
        # Save a reference to existing instance variables that can be ignored by the renderer
      end
    end
  end
end
```

This would ensure:

1. The code we need runs after all gems have included their own `before_action`s (at least theoretically... I think we'd need to test this to verify)
2. The application developer explicitly opts into the behavior
3. The application developer has final control (and responsibility) of the filter order

buhrmi:
Maybe we can override or monkey-patch https://github.com/rails/rails/blob/master/actionpack/lib/abstract_controller/base.rb#L219 to inject some code that is guaranteed to be executed **after** all before_filters have been processed.

BrandonShar:
We could make that before and the opt in a single step. Maybe a module like `include InertiaInstanceProps` or a class method like `inertia_instance_props!` could handle setting up the before filter and activating the props feature on a per request basis.

bknoles:
@BrandonShar yeah, the idea is to treat the `append_before_action` as an opt-in. 

I originally wrote the example code as a class method (just like `inertia_instance_props!` as you suggested), but since the `before_action` order matters a lot i thought it made more sense not to hide that from the developer.

In the end, most devs are probably just going to copy-paste the line from the docs, so making it explicit that this sets up a `before_action` filter that needs to run last could help with debugging, should someone get some extra unwanted props showing up in their frontend templates.

bknoles:
@buhrmi it's not a real Rails gem unless we are monkey patching some part of the core classes, is it? 😜 

(i'd prefer to find another route, but i do love all the ideas being passed around on this feature!)

buhrmi:
I do like @BrandonShar 's idea to have a controller class method like `inertia_instance_props`. When called, it should capture all instance variables that have been created up to that point, and exclude those from being being converted to props. This way, a developer can easily change which instance variables are converted, by changing the position of the `inertia_instance_props` call among the other before_filters. I think that this behavior would be very intuitive to document and understand.

BrandonShar:
Per all of this great feedback, I made the following changes:

1. Implicit props are opt-in only via a class method called `use_inertia_instance_props`. Adding this method to your controller sets up a `before_filter` that enables instance props AND logs all of the instance props defined up to that point and removes them from the inertia render. This may not be 100% in all cases, but it gives control to the user for where its defined.
2. When ANY inertia props are explicitly passed via the normal render function, implicit props are disabled instead of being merged. This will allow users to opt-out on a per response basis if need be.
3. The `inertia_render` function was removed in favor of enabling `render inertia: true` syntax to handle the same thing.

lib/inertia_rails/controller.rb
```diff
@@ -20,6 +20,26 @@ def inertia_share(**args, &block)
           InertiaRails.share_block(block) if block
         end
       end
+
+      def inertia_instance_props
```
bknoles, line 24: Is this method leftover from an earlier iteration? I don't see it called anywhere else in the PR

bknoles:
I like the per-controller opt-in and pre-request opt out. And `render inertia: true` looks so great @ledermann !

lib/inertia_rails/controller.rb
BrandonShar: thank you, great catch!

BrandonShar:
A nice side effect here too is that a user could override `inertia_view_assigns` if they want more fine tuned control over what implicitly makes it way into the renderer.

PR #84, opened 2022-07-23
include initial page data in server-rendered page

buhrmi:
Currently, the Rails adapter throws away the initial page data in the server-side rendered HTML. This makes it impossible for the client to "take over" and hydrate the page. This PR keeps the initial data and also adds a `data-server-rendered="true"` attribute to it, bringing it in line with the other adatpers.

### DIFF
--- a/app/views/inertia.html.erb
+++ b/app/views/inertia.html.erb
@@ -1,1 +1,1 @@
-<div id="app" data-page="<%= page.to_json %>"></div>
+<div <% if ssr_html %> data-server-rendered="true" <% end %>id="app" data-page="<%= page.to_json %>"><%= ssr_html %></div>
--- a/lib/inertia_rails/renderer.rb
+++ b/lib/inertia_rails/renderer.rb
@@ -22,8 +22,8 @@ module InertiaRails
         @response.set_header('X-Inertia', 'true')
         @render_method.call json: page, status: @response.status, content_type: Mime[:json]
       else
-        return render_ssr if ::InertiaRails.ssr_enabled? rescue nil
-        @render_method.call template: 'inertia', layout: ::InertiaRails.layout, locals: (view_data).merge({page: page})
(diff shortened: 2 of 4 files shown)


BrandonShar:
Hey @buhrmi thanks for the PR. The reason Inertia doesn't render the template is because it's returning the already rendered HTML response from the internal node server. The process basically goes like this when SSR is enabled: 
1. Inertia sends a request to the node server with the props that would be rendered
2. the node server fully renders an html page and returns it 
3. Inertia returns the fully rendered html page as the response. 
 
Because we're returning the full html and javascript isn't involved, there's no hydration process on the initial load of an SSR page. If the user has javascript enabled, regular inertia requests will take over afterwards.

Let me know if my explanation doesn't make sense!

buhrmi:
> If the user has javascript enabled, regular inertia requests will take over afterwards.

How? How can inertia take over, if the template isn't rendered and the `<div id="app" data-page="...">` doesn't exist in the HTML that is sent to the browser on first load? Where would inertia mount the app?

BrandonShar:
I just ran a Vite/React SSR build locally and the app div with the data-page is rendered on the page with all of the SSR rendered html within it.

Here's the rendered HTML from my [inertia template repo](https://github.com/BrandonShar/inertia-rails-template)
[inertia-rails-template.txt](https://github.com/inertiajs/inertia-rails/files/9174337/inertia-rails-template.txt)

buhrmi:
Yeah, I just looked at the source of https://pingcrm-vite.herokuapp.com/login (from @ElMassimo ). It also includes the `<div data-server-rendered="true" id="app" data-page=...`. I can't figure out where it comes from because I can't find the string "server-rendered" anywhere in the inertia-rails repo 🤔🤔

buhrmi:
Okay, after digging a little bit deeper, it looks like the client-side adapters are responsible for rendering the intial `<div id="app">` element. Will close this PR.

PR #86, opened 2022-08-05
feat: support multiple inertia apps by allowing different layouts

ElMassimo:
### Description 📖 

This pull request allows to leverage the default layout resolution in Rails.

### Background 📜 

When using Rails as the backend for different Inertia apps, having a global layout configuration does not allow each app to have a different layout (with different title and meta tags, for example).

By using the layout specified by the controller (which defaults to `application` in most apps), the user has more flexibility to choose which layout should be used.

---

For example, this pull request enables:

```ruby
class Admin::BaseController < ApplicationController
  layout "admin"
end
```

where the `admin` frontend app gets its own layout, with different scripts, meta tags, and title.

### Notes ✏️ 

- The previous mechanism to configure a layout is still supported:

  ```ruby
  InertiaRails.configure { |config| config.layout = 'other' }
  ```

- An internal `inertia_layout` controller instance method was added for the purpose of even more dynamic configurations, for users that need the flexibility. Now they can customize the behavior by defining a method in their controllers, without having to monkey-patch `InertiaRails`.

### Considerations about Backwards Compatibility

This shouldn't be considered a breaking change, as the default layout for most Rails apps is `application` (hence, the tests didn't need to be updated), and in other setups, a user would already be configuring `InertiaRails::Configuration#layout` explicitly, which will continue to work.

### DIFF
--- a/lib/inertia_rails/controller.rb
+++ b/lib/inertia_rails/controller.rb
@@ -40,6 +40,9 @@ module InertiaRails
 
     def inertia_layout
       layout = ::InertiaRails.layout
+
+      # When the global configuration is not set, let Rails decide which layout
+      # should be used based on the controller configuration.
       layout.nil? ? true : layout
     end
 
--- a/lib/inertia_rails/controller.rb
+++ b/lib/inertia_rails/controller.rb
@@ -38,6 +38,11 @@ module InertiaRails
 
     private
 
+    def inertia_layout
+      layout = ::InertiaRails.layout
+      layout.nil? ? true : layout
+    end
+
     def inertia_location(url)
       headers['X-Inertia-Location'] = url
       head :conflict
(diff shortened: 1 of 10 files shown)


lib/inertia_rails/controller.rb
```diff
@@ -38,6 +38,11 @@ def redirect_back(fallback_location:, allow_other_host: true, **options)
 
     private
 
+    def inertia_layout
+      layout = ::InertiaRails.layout
+      layout.nil? ? true : layout
```
BrandonShar, line 43: What does returning `true` do here?

ElMassimo: I'll add an inline doc, it lets Rails know that the view should use a layout.

BrandonShar (approved):
:+1:

PR #94, opened 2023-03-10
Fix rspec failure on X-Inertia requests

iatanas0v:
As initial request and subsequent requests are handled with different renderers sometimes we may end up in situation where controller action may fail just in one of those cases:

https://github.com/inertiajs/inertia-rails/blob/master/lib/inertia_rails/renderer.rb#L19L28

This is exactly what happened on one of my work projects. Subsequent requests (with `X-Inertia` header were failing) but my rspec specs were all :green_circle: .

Naturally, I decided to add another spec that issues the request with `X-Inertia` header. It turned out that the rspec helper is not expecting that :sweat_smile: . This PR aims to resolve that and allow tests like the one below.

```ruby
it 'works on subsequential requests' do
  get homepage_path, headers: {'X-Inertia' => true}

  expect(inertia_props.keys).to include('records', 'pagination', 'filters')
end
```

### DIFF
--- a/lib/inertia_rails/rspec.rb
+++ b/lib/inertia_rails/rspec.rb
@@ -25,9 +25,16 @@ module InertiaRails
       protected
     
       def set_values(params)
-        @view_data = params[:locals].except(:page)
-        @props = params[:locals][:page][:props]
-        @component = params[:locals][:page][:component]
+        if params[:locals].present?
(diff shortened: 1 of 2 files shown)


BrandonShar (approved):
Thanks @iatanas0v and sorry for taking so long to review this, I"m not sure how I missed it!

PR #95, opened 2023-04-29
Added support for Vite in installer

### LINKED ISSUE #85: Vite instead of Webpack
Looking at the generators I was checking that the vue installer only works for webpacker apps, which in Rails 7 is not gonna be the main way to go. Will this be updated anytime soon?

wJoenn:
I figured it wouldn't take too long so in the end I did a first draft today.
I have no idea how to test this in my local environment so it has not been tested, sorry about that.
If you deem it necessary feel free to give me a hint on the procedure and I'll do that

The idea is simple, we check if vite is installed and we store that boolean in a `@vite_installed?` variable.
We then use that variable to render paths accordingly

The `vite_source_code_dir` method is used to read the `config/vite.json` file and grab path to the sourceDir there in case some users have changed that path in their config. It also verifies whether the path includes a closing "/" or not and adds it it's not the case. Again that is to unsure that it wouldn't conflict with different user's config

I created an instances variables instead of methods for `@vite_installed?` and `@vite_source_code_dir` because I wanted to avoid reading the file or executing `rails vite:verify_install` multiple times.

Closes #85

### DIFF
--- a/lib/generators/inertia_rails/install_generator.rb
+++ b/lib/generators/inertia_rails/install_generator.rb
@@ -12,6 +12,7 @@ module InertiaRails
     def install
       exit! unless installable?
 
+      @vite_source_code_dir = vite_source_code_dir if @vite_installed?
       install_base!
 
       send "install_#{options[:front_end]}!"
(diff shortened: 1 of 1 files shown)


wJoenn:
Just saying but I noticed the installer is still using old inertia packages instead of the ones introduce with [Inertia 1.0](https://inertiajs.com/upgrade-guide)
I can try to update that if you want

PR #96, opened 2023-06-26
Set up Rails CSRF to play nice with Axios default CSRF behavior

### LINKED ISSUE #71: Making CSRF protection work out of the box

bknoles:
Resolves #71

More discussion there and #72

### DIFF
--- a/lib/generators/inertia_rails/install/react/inertia.jsx
+++ b/lib/generators/inertia_rails/install/react/inertia.jsx
@@ -1,15 +1,11 @@
 import { App } from '@inertiajs/inertia-react';
 import React from 'react';
 import { render } from 'react-dom';
-import axios from 'axios';
 import { InertiaProgress } from '@inertiajs/progress';
 
 document.addEventListener('DOMContentLoaded', () => {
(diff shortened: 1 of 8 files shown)


PedroAugustoRamalhoDuarte:
It's an excellent feature that enables quick setup of Inertia client-side following the website https://inertiajs.com/client-side-setup

PR #102, opened 2023-08-22
Fix partial reload rendering after deep merge changes

bknoles:
This fixes a regression introduced by #99 .

It snuck past our test suite because the error happened on both sides of the expectation.

### DIFF
--- a/lib/inertia_rails/renderer.rb
+++ b/lib/inertia_rails/renderer.rb
@@ -45,7 +45,7 @@ module InertiaRails
     def computed_props
       _props = ::InertiaRails.shared_data(@controller).send(prop_merge_method, @props).select do |key, prop|
         if rendering_partial_component?
-          key.in? partial_keys
+          key.to_sym.in? partial_keys
         else
           !prop.is_a?(InertiaRails::Lazy)
         end
--- a/spec/inertia/rendering_spec.rb
+++ b/spec/inertia/rendering_spec.rb
@@ -64,7 +64,7 @@ RSpec.describe 'rendering inertia views', type: :request do
 
   context 'partial rendering' do
     let (:page) {
-      InertiaRails::Renderer.new('TestComponent', controller, request, response, '', props: { sport: 'hockey'}).send(:page)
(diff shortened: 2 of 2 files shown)


BrandonShar (approved):

lib/inertia_rails/renderer.rb
```diff
@@ -45,7 +45,7 @@ def layout
     def computed_props
       _props = ::InertiaRails.shared_data(@controller).send(prop_merge_method, @props).select do |key, prop|
         if rendering_partial_component?
-          key.in? partial_keys
+          key.to_sym.in? partial_keys
```
bknoles, line 48: `partial_keys` are cast to symbols. In #99 , we changed `@props` to be a hash with indifferent access. That casts the keys of `@props` to strings. Thus, `computed_props` was filtering every single prop out, completely breaking partial reloads.

spec/inertia/rendering_spec.rb
```diff
@@ -64,7 +64,7 @@
 
   context 'partial rendering' do
     let (:page) {
-      InertiaRails::Renderer.new('TestComponent', controller, request, response, '', props: { sport: 'hockey'}).send(:page)
+      InertiaRails::Renderer.new('TestComponent', controller, request, response, '', props: { sport: 'hockey' }).send(:page)
```
bknoles, line 67: The sneaky part is here: `.page` is compared to `response.body`, which is the subject of the test file. This test happily passed because both `.page` and `response.body` were filtering out all the props, leaving a comparison between `{ props: {} }` and  `{ props: {} }`.

It failed once modified to check for the actual content we wish to see.

spec/inertia/rendering_spec.rb
```diff
@@ -107,13 +108,16 @@
...
+        'X-Inertia-Partial-Data' => 'sport,level',
         'X-Inertia-Partial-Component' => 'TestComponent',
       }}
 
       before { get lazy_props_path, headers: headers }
 
       it { is_expected.to eq page.to_json }
+      it { is_expected.to include('basketball') }
+      it { is_expected.to include('worse') }
+      it { is_expected.not_to include('intense') }
```
bknoles, line 120: A more explicit spec here should help us catch future regressions.

PR #106, opened 2023-09-28
add railties as the runtime dependency

BenMorganMY:
This releases this gem from depending on the whole Rails suite.

### DIFF
--- a/inertia_rails.gemspec
+++ b/inertia_rails.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.add_runtime_dependency "rails", '>= 5'
+  spec.add_runtime_dependency "railties", '>= 5'
 
   spec.add_development_dependency "bundler", "~> 2.0"
   spec.add_development_dependency "rake", "~> 13.0"


BrandonShar:
Thanks @BenMorganMY ! Could you explain a bit more about how we would benefit from this change?

Is there anything we lose by no longer requiring rails?

BenMorganMY:
@BrandonShar no, inertia doesn't lose anything from it. Some rails apps actually don't depend on rails itself. For example, we only pull in the rails gems that we need and leave out `actiontext`, `activestorage`, `actioncable`, and so on. Inertia's dependency was very broad and was pulling in gems that it doesn't actually rely on or use.

BenMorganMY:
Hey @BrandonShar, just pinging to see if you're able to get this merged this week and a new release cut?

BrandonShar:
@bknoles what do you think?

I think it probably makes the most sense, my only super minor hedge is what if we do want to depend on one of those rails gems in the future?

BenMorganMY:
@BrandonShar if you need to depend on one of the rails gems in the future, you would want to add it as a dependency in your gemspec.

BrandonShar:
Oh duh... I'm thinking about a future breaking change, but there wouldn't be one at all; the gem still owns the dependency. Thanks I'll get it merged in!

BrandonShar (approved):

BenMorganMY:
Awesome. Let me know when a new gem is cut.

bknoles:
v3.1.3 hot off the presses!

https://rubygems.org/gems/inertia_rails/versions/3.1.3

I'm chiming in late, but no retroactive objections to merging this. Tests all pass, and I'm happy to have a smaller footprint!

PR #111, opened 2024-04-11
Refactor inertia share to use instance variables

PedroAugustoRamalhoDuarte:
## Goal

This PR refactors how `inertia_share` is storage inside the request lifecycle.

## Solution

The solution was to sticky with default rails conventions, using only instance variable to storage inertia_share data and using inertia_share as a wrapper to rails before_action

## Future improves

- Maybe we can support easly `only`, `expect` features  from [before_action](https://api.rubyonrails.org/v7.1.3.2/classes/AbstractController/Callbacks/ClassMethods.html#method-i-_insert_callbacks).


buhrmi:
wow that was fast

PedroAugustoRamalhoDuarte:
@bknoles i see the problem with class_attribute and i refactor with this idea in mind: https://github.com/inertiajs/inertia-rails/pull/108#issuecomment-2050573597.

Basically using instance attributes and `before_action`, the CI is passing, but i think we need some more tests

lib/inertia_rails/controller.rb
```diff
@@ -41,6 +63,11 @@ def default_render
       end
     end
 
+    def shared_data
+      return {} unless @_inertia_shared_plain_data
```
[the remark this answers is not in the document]
PedroAugustoRamalhoDuarte: Yes, you're absolutely correct.

lib/inertia_rails/controller.rb
```diff
@@ -31,6 +41,18 @@ def use_inertia_instance_props
           @_inertia_skip_props = view_assigns.keys + ['_inertia_skip_props']
```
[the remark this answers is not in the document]
PedroAugustoRamalhoDuarte: I don't think so, the other before action that initialize these variables will always run anyway inside controller

lib/inertia_rails/controller.rb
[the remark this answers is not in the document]
PedroAugustoRamalhoDuarte: @buhrmi I think we need more attention in this method. There is this safeguard only for if rendered method is called without a controller. Because in the controller will always setup the initial value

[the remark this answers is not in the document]
PedroAugustoRamalhoDuarte: I dont think is worth to make something like this
```suggestion
            return {} if @_inertia_shared_plain_data.nil? && @_inertia_shared_blocks.nil?
      return evaluated_blocks if if @_inertia_shared_plain_data.nil?
```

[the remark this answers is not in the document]
PedroAugustoRamalhoDuarte: I have made a refactor in this one

bknoles:
Want to review the code a bit more, but, as I mentioned in #115 , I'm leaning towards merging this!

lib/inertia_rails/controller.rb
```diff
@@ -31,6 +41,22 @@ def use_inertia_instance_props
...
         end
       end
+
+      def share_plain_data(hash)
+        @_inertia_shared_plain_data = @_inertia_shared_plain_data.merge(hash)
+      end
+
+      def share_block(&block)
+        @_inertia_shared_blocks = @_inertia_shared_blocks + [block]
+      end
```
PedroAugustoRamalhoDuarte, line 51: we really need this two methods? maybe we can remove then

spec/inertia/conditional_sharing_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+      normal_shared_prop: 1,
+      show_only_prop: 1,
+      conditionally_shared_show_prop: 1,
+    })
+
+    get conditional_share_index_path, headers: {'X-Inertia' => true}
+    expect(JSON.parse(response.body)['props'].deep_symbolize_keys).to eq({
+      index_only_prop: 1,
+      normal_shared_prop: 1,
+      conditionally_shared_show_prop: 1, # TODO: Because we call before_action twice, this is leaking.
```
PedroAugustoRamalhoDuarte, line 24: @bknoles i have cherry picked this failling spec, for now i add a TODO untill we think about

PedroAugustoRamalhoDuarte:
The tests that are failling in CI are bit flaky, i dont know if is a error in the test code or in the new controller implementation i can check better in this friday

spec/inertia/conditional_sharing_spec.rb
PedroAugustoRamalhoDuarte: In this one i think the best solution is already implement one of this approach to run the inertia_share to only some controller method https://github.com/inertiajs/inertia-rails/issues/115#issuecomment-2081728942.

`before_actions_options` is a good explanatory name, but i think maybe is too big i would go for only `options` or `share_options` to make less conflict  with existing code

waliffcordeiro:
Great addition! Refactoring to stick with default Rails conventions by using instance variables to store `inertia_share` data and using `inertia_share` as a wrapper for `before_action` improves the clarity and maintainability of the code. Future enhancements, like supporting `only` and `except` features from `before_action`, would also be beneficial and add more flexibility. Good job!

lib/inertia_rails/controller.rb
```diff
@@ -41,6 +67,14 @@ def default_render
       end
     end
 
+    def shared_data
+      # Theses variables are set in the before_action, but we need to ensure they are initialized when using renderer alone
+      @_inertia_shared_plain_data ||= {}
+      @_inertia_shared_blocks ||= []
+
+      @_inertia_shared_plain_data.merge(evaluated_blocks)
```
bknoles, line 75: This is a nitpick, but I'd prefer not to set instance variables inside a reader-like method when the method has a different name than the instance variable.

I'd prefer one of two options:

Create explicit reader methods that hold the defaults (and change the references in the controller to use this method instead of accessing the instance variable directly)

```ruby
module InertiaRails
  module Controller
    def inertia_shared_plain_data
      @_inertia_shared_plain_data ||= {}
    end

    def inertia_shared_blocks
      @_inertia_shared_blocks ||= []
    end
  end
end
```

or, just put the defaults inline in the reader methods:

```ruby
module InertiaRails
  module Controller
    def shared_data
      (@_inertia_shared_plain_data || {}).merge(evaluated_blocks)
    end

    def evaluated_blocks
      def evaluated_blocks
        (@_inertia_shared_blocks || []).map { |block| instance_exec(&block) }&.reduce(&:merge) || {}
      end
    end
  end
end
```
 
I don't have a strong opinion on which one.

lib/inertia_rails/inertia_rails.rb
```diff
@@ -3,20 +3,10 @@
 require 'inertia_rails/lazy'
 
 module InertiaRails
-  thread_mattr_accessor :threadsafe_shared_plain_data
-  thread_mattr_accessor :threadsafe_shared_blocks
-  thread_mattr_accessor :threadsafe_html_headers
-
```
bknoles, line 9: goodbye, fickle threadsafe class variables!

spec/inertia/sharing_spec.rb
```diff
@@ -54,52 +54,57 @@
     it { is_expected.to eq props.merge({ errors: errors }) }
   end
 
-  context 'multithreaded intertia share' do
+  context 'when multithreaded with shared data' do
```
bknoles, line 57: Could you undo the change to this spec? The updated code should be threadsafe just by design (making this spec unncessary), but I'd prefer to test for a regression against the original spec.

spec/inertia/ssr_spec.rb
```diff
@@ -1,7 +1,8 @@
+require 'net/http'
```
bknoles, line 1: Genuine question, what made this necessary?

bknoles:
> Great addition! Refactoring to stick with default Rails conventions by using instance variables to store `inertia_share` data and using `inertia_share` as a wrapper for `before_action` improves the clarity and maintainability of the code. Future enhancements, like supporting `only` and `except` features from `before_action`, would also be beneficial and add more flexibility. Good job!

Team support! I like it! 😀

spec/inertia/conditional_sharing_spec.rb
PedroAugustoRamalhoDuarte: Good ideia i will keep with the test

spec/inertia/sharing_spec.rb
PedroAugustoRamalhoDuarte: In the old implementation there was this comment:

        # Would prefer to make this a second get request, but RSpec will overwrite
        # the @response variable if another request is made in the second thread.
        # This simulates the relevant effects of another call to a different
        # controller with different values for inertia_share
        
  But i refactor to make another request and it working fine

bknoles (approved):
At long last, a merge! Thanks for sticking with it!

bknoles:
Released as part of `3.2.0`!

PR #116, opened 2024-04-29
Conditional shared data with instance variables

bknoles:


bknoles:
Closing since this was only part of an edge case discussion for #111

PR #117, opened 2024-04-29
Class attribute failure

bknoles:


bknoles:
closing since this was just an edge case example for the discussion in #111

PR #118, opened 2024-05-04
Set the correct `Vary` header

osbre:
~Following Inertia author on this one: https://github.com/inertiajs/inertia-laravel/pull/404, since he updated the `Vary` header to be `X-Inertia` instead of `Accept`.~

**The [protocol](https://inertiajs.com/the-protocol#inertia-responses) got updated and now `Vary` has to be `X-Inertia`.**

### DIFF
--- a/lib/inertia_rails/renderer.rb
+++ b/lib/inertia_rails/renderer.rb
@@ -19,7 +19,7 @@ module InertiaRails
 
     def render
       if @request.headers['X-Inertia']
-        @response.set_header('Vary', 'Accept')
+        @response.set_header('Vary', 'X-Inertia')
         @response.set_header('X-Inertia', 'true')
         @render_method.call json: page, status: @response.status, content_type: Mime[:json]
       else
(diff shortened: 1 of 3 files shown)


osbre:
@bknoles The [protocol](https://inertiajs.com/the-protocol#inertia-responses) page got updated, so now it's part of a standard. I think having this change merged benefits the community.

bknoles:
Looks good to me! I see you've [been active](https://github.com/inertiajs/inertia-phoenix/pull/4) on this one @osbre ! Thanks!

For anyone curious what's going on here (I had to do some research myself), the Vary header allows things like browser caches to differentiate between otherwise identical looking requests. In Inertia, the full page HTML response vs. the Inertia-fied XHR response are mostly the same, except that the XHR request includes the `X-Inertia: true` header. The initial version of the Inertia adapters set a header "Vary: Accept". This actually did allow browsers to differentiate between the HTML and XHR requests, but only because of some browser quirks. This change makes the Vary header behavior more robust and less dependent on the browser implementation.

bknoles:
Released as part of `3.2.0`!

osbre:
Cool! Also, releases aren't synced with tags and it might make a misleading impression that the latest release was in 2019. IMO, it is best to hide this part (see the icon button from the 2nd screenshot) or keep it up to date.



image



image

bknoles:
Done! That was an artifact of originally publishing via Github. Thanks for the call out.

PR #121, opened 2024-06-26
feat: add `inertia_config` to enable controller-specific config

ElMassimo:
### Description 📖 

This pull request is a complete refactor of the implementation in order to introduce controller-specific configuration.

### `inertia_config`

A new way for controllers to override the global configuration, and provide dynamic values, such as:

```ruby
  inertia_config(
    ssr_enabled: -> { action_name == "index" },
  )
end
```

Just like `inertia_share`, configuration is inheritable, and subclasses can override values.

In addition, all configuration options can now be procs (before only `version` supported this).

### Performance 🚀 

The implementation of `inertia_share` no longer relies on `before_action`, improving performance when calculating shared data at runtime, and reducing object allocations.

As an additional benefit, we no longer run this code unnecessarily for non-Inertia requests.

### Notes ✏️ 

Each commit is a gradual refactoring, and can be read in isolation (and passes all tests). I'd still suggest using _Squash and Merge_.

We also `freeze` config and shared data to prevent any mutations, ensuring it's safe for multiple threads to access the configuration from the controller class.

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,7 +191,7 @@ inertia 'about' => 'AboutComponent'
 
 Enable SSR via the config settings for `ssr_enabled` and `ssr_url`.
 
-When using SSR, don't forget to add `<%= inertia_headers %>` to the `<head>` of your `application.html.erb`.
+When using SSR, don't forget to add `<%= inertia_ssr_head %>` to the `<head>` of your `application.html.erb`.
 
 ## Configuration
 
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ end
 In order to use instance props, you must call `use_inertia_instance_props` on the controller (or a base controller it inherits from). If any props are provided manually, instance props
 are automatically disabled for that response. Instance props are only included if they are defined after the before filter is set from `use_inertia_instance_props`.
 
-Automatic component name is also opt in, you must set the `default_render` config value to `true`. Otherwise, you can simply `render inertia: true` for the same behavior explicitly. 
+Automatic component name is also opt in, you must set the [`default_render`](#default_render) config value to `true`. Otherwise, you can simply `render inertia: true` for the same behavior explicitly.
 
 ### Layout 
 
@@ -132,7 +132,7 @@ end
 }
 ```
 
-Deep merging can be configured using the `deep_merge_shared_data` configuration option.
+Deep merging can be configured using the [`deep_merge_shared_data`](#deep_merge_shared_data) configuration option.
 
 If deep merging is enabled, you can still opt-out within the action:
 
@@ -182,7 +182,7 @@ inertia 'about' => 'AboutComponent'
 
 ### SSR _(experimental)_
 
-Enable SSR via the configuration options for `ssr_enabled` and `ssr_url`.
+Enable SSR via the configuration options for [`ssr_enabled`](#ssr_enabled-experimental) and [`ssr_url`](#ssr_url-experimental).
 
 When using SSR, don't forget to add `<%= inertia_ssr_head %>` to the `<head>` of your layout (i.e. `application.html.erb`).
 
--- a/README.md
+++ b/README.md
@@ -7,14 +7,25 @@
 
 ### Backend
 
-Just add the inertia rails gem to your Gemfile
+Add the `inertia_rails` gem to your Gemfile.
+
 ```ruby
 gem 'inertia_rails'
 ```
 
+For more instructions, see [Server-side setup](https://inertia-rails.netlify.app/guide/server-side-setup.html).
+
 ### Frontend
 
-Rails 7 specific frontend docs coming soon. For now, check out the official Inertia docs at https://inertiajs.com/ or see an example using React/Vite [here](https://github.com/BrandonShar/inertia-rails-template)
+We are discussing on bringing official docs for Inertia Rails to this repo, as
+the [official docs](https://inertiajs.com/client-side-setup) are specific to Laravel.
+
+In the meantime, you can refer to the community-maintained [Client-side setup](https://inertia-rails.netlify.app/guide/client-side-setup.html).
+
+Examples:
+
+- [React/Vite](https://github.com/BrandonShar/inertia-rails-template)
+- [React/Vite + SSR](https://github.com/ElMassimo/inertia-rails-ssr-template)
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -67,8 +67,7 @@ Automatic component name is also opt in, you must set the `default_render` confi
 
 ### Layout 
 
-Inertia layouts use the rails layout convention and can be set or changed in the same way. The original `layout` config option is still functional, but will likely be deprecated in the future in favor
-of using rails layouts.
+Inertia layouts use the rails layout convention and can be set or changed in the same way.
 
 ```ruby
 class EventsController < ApplicationController
@@ -133,20 +132,14 @@ end
 }
 ```
 
-Deep merging can be set as the project wide default via the InertiaRails configuration:
+Deep merging can be configured using the `deep_merge_shared_data` configuration option.
 
-```ruby
-# config/initializers/some_initializer.rb
-InertiaRails.configure do |config|
-  config.deep_merge_shared_data = true
-end
-
-```
-
-If deep merging is enabled by default, it's possible to opt out within the action:
+If deep merging is enabled, you can still opt-out within the action:
 
 ```ruby
 class CrazyScorersController < ApplicationController
+  inertia_config(deep_merge_shared_data: true)
+
   inertia_share do
     {
       basketball_data: {
@@ -163,7 +156,7 @@ class CrazyScorersController < ApplicationController
   end
 end
 
-# Even if deep merging is set by default, since the renderer has `deep_merge: false`, it will send a shallow merge to the frontend:
+# `deep_merge: false` overrides the default:
 {
   basketball_data: {
     points: 100,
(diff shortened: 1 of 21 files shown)


lib/inertia_rails/middleware.rb
```diff
@@ -71,17 +74,15 @@ def inertia_request?
...
-
-        InertiaRails.version.is_a?(Numeric) ? inertia_version.to_f : inertia_version
+      def server_version
+        controller.send(:inertia_configuration).version
       end
 
-      def saved_version
-        InertiaRails.version.is_a?(Numeric) ? InertiaRails.version.to_f : InertiaRails.version
+      def coerce_version(version)
+        server_version.is_a?(Numeric) ? version.to_f : version
```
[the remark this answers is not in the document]
ElMassimo: No, that's the original logic.

lib/inertia_rails/configuration.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+module InertiaRails
+  class Configuration
+    DEFAULTS = {
+      default_render: false,
+
+      # Let Rails decide which layout should be used based on the controller configuration.
+      layout: true,
```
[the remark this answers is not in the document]
ElMassimo: I made the original change in #86.

The controller method was returning `true` if there global was `nil`, so the behavior stays the same.

spec/dummy/app/views/layouts/application.html.erb
```diff
@@ -1,3 +1,3 @@
-<%= inertia_headers %>
+<%= local_assigns[:inertia_ssr_head] %>
```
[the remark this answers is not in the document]
ElMassimo: Agreed 👍 

I would add it with a warning that it will be removed.

These are not headers, it's `html` for the `head`, so the name is inaccurate.

In addition, it does not make sense to call this in the controller, as it wouldn't have been set yet.

lib/inertia_rails/configuration.rb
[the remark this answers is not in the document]
skryukov: Just tested, if someone sets `nil` manually, this PR will brake their code (Rails renders page without layout):

```
  config.layout = nil
end
```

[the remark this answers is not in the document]
ElMassimo: It's unlikely that users are doing that, given it's the default.

That behavior would also be consistent with using `layout: false`.

If this is released in a major version, I'd say it's more intuitive if whatever the user configures is not overriden internally.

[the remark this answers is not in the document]
skryukov: yup, that's a great addition, and since README uses `true` as a default, a note in the changelog should be enough, I guess 👍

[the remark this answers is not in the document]
ElMassimo: I don't feel strongly about it, we could be strict in backwards compatibility first, and remove it later, just mentioning that I don't think it's as useful a setting after #86.

Thanks for the review 😃

spec/inertia/sharing_spec.rb
```diff
@@ -54,63 +54,6 @@
     it { is_expected.to eq props.merge({ errors: errors }) }
   end
 
-  context 'multithreaded inertia share' do
```
PedroAugustoRamalhoDuarte, line 57: @ElMassimo just for curiosity, why do you remove this test? It was failing or because we do not have this problem anymore by default?

PedroAugustoRamalhoDuarte (commented):
@ElMassimo Excellent PR, very important to make easy to set up Inertia V2. This #131 task could be easily resolved if this PR is merged

spec/inertia/sharing_spec.rb
ElMassimo: It's not possible to have multi-threading issues anymore, as the objects that can be shared between threads are now frozen (not writable, and thus not susceptible to concurrent writes either).

PedroAugustoRamalhoDuarte: Thanks for the response!!

lib/inertia_rails/controller.rb
```diff
@@ -77,19 +79,27 @@ def redirect_back(fallback_location:, allow_other_host: true, **options)
...
 
-    private
+    def inertia_configuration
+      self.class._inertia_configuration.bind_controller(self)
+    end
 
-    def inertia_layout
-      layout = ::InertiaRails.layout
+    def inertia_shared_data
+      initial_data = session[:inertia_errors].present? ? {errors: session[:inertia_errors]} : {}
```
skryukov, line 94: I used to set `inertia_share(errors: {})` to mimic Laravel plugin's default, with this change — I always overwrite inertia's errors with my empty hash.

Should we move this to the renderer and apply at the end?

lib/inertia_rails/controller.rb
```diff
@@ -1,37 +1,32 @@
...
+        cookies['XSRF-TOKEN'] = form_authenticity_token if protect_against_forgery?
       end
     end
 
     module ClassMethods
-      def inertia_share(hash = nil, &block)
-        before_action do
-          @_inertia_shared_plain_data = @_inertia_shared_plain_data.merge(hash) if hash
-          @_inertia_shared_blocks = @_inertia_shared_blocks + [block] if block_given?
+      def inertia_share(**attrs, &block)
```
skryukov, line 17: This is a breaking change, should we continue supporting hashes or at least throw a deprecation at first, wdyt?

```ruby
inertia_share({foo: 'bar'}) # => throws error
```

lib/inertia_rails/middleware.rb
```diff
@@ -71,17 +74,15 @@ def inertia_request?
...
-        sent_version != saved_version
+        coerce_version(client_version) != coerce_version(server_version)
       end
 
-      def sent_version
-        return nil if inertia_version.nil?
-
-        InertiaRails.version.is_a?(Numeric) ? inertia_version.to_f : inertia_version
+      def server_version
+        controller.send(:inertia_configuration).version
```
skryukov, line 81: Controller might be nil (i.e. in case of 404 error), I think this is a valid fix:

```suggestion
        config = controller ? controller.send(:inertia_configuration) : ::InertiaRails.configuration
        config.version
```

bknoles:
Hey @ElMassimo thanks for this PR. I'm digging into it now (at long last)!

Just as a heads up, I resolved a merge conflict in the README and committed it to your branch. Hope that's not stepping on your toes.

spec/inertia/sharing_spec.rb
bknoles: Fwiw, i pulled the branch and added these tests back, and they do still pass.

These were made unnecessary by #111. At the time, I asked @PedroAugustoRamalhoDuarte to keep the specs so we weren't changing spec + implementation at the same time. I'm good with removing them now that we have evidence that refactor passed the original spec. (kind of like adding a new DB column, migrating data, then removing the old column once everything is verified working in production instead of just renaming the column in one shot)

lib/inertia_rails/controller.rb
bknoles: I think what you are describing is only working by accident. Before [this change](https://github.com/inertiajs/inertia-rails/pull/111/files#diff-d3e9c2c338687f7af19e477ad05c7dfae9ee7e5ae2fb87adb72f9694b143bc87L11), I believe `inertia_share(errors: {})` would have overwritten the errors from the session (exactly as you described). That change moved the error sharing into a block, which gets evaluated _after_ the plain data is computed, allowing your technique to work. I'd call that unintended behavior because presumably this equivalent looking code just wouldn't work:

```ruby
# This will always overwrite the session errors because this block will be evaluated after the error sharing block from the library
inertia_share do
  {errors: {}}
end
```

IMO, if the application code explicitly sets a prop it _should_ overwrite whatever came out of the library internals, so this PR is actually fixing that.

That said, maybe our default should match the Laravel default and always return something in an errors prop? That seems to be the real thing at issue here. I don't have a strong opinion, but I'd lean towards keeping the default we have since it'd be a potentially breaking change.

I consider that separate from the scope of this PR and won't hold up merging it for this.

(Please let me know if I'm misunderstanding your use case)

skryukov: Oh, great, I just noticed a breaking change, although it appears to be a fix 😂.

Regarding `errors: {}` being the default—that's a significant breaking change, and I would suggest adding a compatibility configuration flag for opting into backward compatibility until the next major release. I also think we should first discuss the possibility of adding this behavior to the protocol. But anyway, that's a separate issue 👍

README.md
```diff
@@ -189,34 +192,85 @@ If you don't need a controller to handle a static component, you can route direc
...
+The default configuration can be found [here](https://github.com/inertiajs/inertia-rails/blob/master/lib/inertia_rails/configuration.rb#L5-L22).
+
+### Local Configuration
+
+Use `inertia_config` in your controllers to override global settings:
+
+```ruby
+class EventsController < ApplicationController
+  inertia_config(
+    version: "events-#{InertiaRails.configuration.version}",
```
bknoles, line 227: Local version configuration seems like a bit of a footgun. In this example, won't you trigger a hard refresh every time you visit this controller from a different controller since those versions will always be mismatched?

All the other configuration options make perfect sense to optionally configure per-controller. Maybe we should exclude `version` (and only `version`) from being locally configured?

ElMassimo: > won't you trigger a hard refresh every time you visit this controller from a different controller since those versions will always be mismatched?

You could have different apps, with entirely different Vite bundles, so that would actually make sense and the hard refresh would "make it work". Linking across apps using Inertia navigations isn't "recommended", but it's nice that it would work out of the box.

Having an exception for `version` seems unnecessary, it's completely optional to use it, and in some cases it's actually useful.

lib/inertia_rails/controller.rb
bknoles, line 17:
```suggestion
      def inertia_share(attrs = {}, &block)
```

lib/inertia_rails/middleware.rb
bknoles: Yep, I confirmed this locally. You'd need to make an Inertia request and also use the Rails default 404 handling, but it does indeed throw an error.

I'm gonna have it just return nil instead of trying to fall back to a global configuration; since we can have multiple versions configured now I think we shouldn't try to guess at the version here.

```suggestion
        controller&.send(:inertia_configuration)&.version
```

bknoles (approved):

bknoles:
Made a couple tweak commits and finally got this thing merged! Thanks so much @ElMassimo ! Lovely code that adds some really nice features.

Thanks also to @skryukov for your sharp eyes to catch a few breaking changes that weren't covered by our test suite. I'm going to follow up with a few specs I created locally to cover those things.

I'll get a release out sometime this week in advance of any Inertia v2.0 work.

bknoles:
Released in 3.3.0!

brianmureithi:
What was the solution for shared data?
`# Base controller
  include InertiaRails::Controller

  rescue_from StandardError, with: :handle_standard_error
  before_action :share_current_user

  private

  def handle_standard_error(exception)
    Sentry.capture_exception(exception)
    raise exception
  end

  def share_current_user
    # Use Devise's user_signed_in? method
    puts "user_signed_in? #{user_signed_in?}"

    if user_signed_in?
      # Use Devise's current_user method
      puts "user_signed_in? #{user_signed_in?}"
      user = current_user

      inertia_share(
        current_user: {
          id: user.id,
          email: user.email,
          first_name: user.first_name,
          last_name: user.last_name
        }
      )
    end
  rescue StandardError => e
    Rails.logger.error "Error sharing current user: #{e.message}"
    Rails.logger.error e.backtrace.join("\n")
  end
end
` says inertia sahre is undefined

PR #128, opened 2024-07-21
Set `Vary` header to `X-Inertia` on HTML responses

### LINKED ISSUE #124: `Vary` header should be included in all Inertia responses
Hi! I've stumbled upon a very specific error when an Inertia page is served from the browser cache:

1. First, HTML page is accessed, response is cached
2. Then, a partial reload on the same URL, expecting a Inertia JSON response (with `X-Inertia` header)
3. Inertia gets served first HTML, errors and shows response inside a modal

Taking a look at the `InertiaRails::Renderer`, it seems that the `Vary` header that would prevent this from happening is only added for subsequent `X-Inertia` Inertia responses – and so the browser still uses the first response for all cache hits.

I'm testing a monkey-patch fix on my app and could open a PR if this is the correct diagnosis.

skryukov:
This PR Fixes #124 by setting the `Vary` header to `X-Inertia` on all Inertia responses.

This change brings Rails adapter in line with the Laravel adapter:
https://github.com/inertiajs/inertia-laravel/blob/ddf47e999dcdabd2f370e5b9772a7449f4a81729/src/Middleware.php#L87

### DIFF
--- a/lib/inertia_rails/renderer.rb
+++ b/lib/inertia_rails/renderer.rb
@@ -16,8 +16,8 @@ module InertiaRails
     end
 
     def render
+      @response.set_header('Vary', 'X-Inertia')
       if @request.headers['X-Inertia']
-        @response.set_header('Vary', 'X-Inertia')
         @response.set_header('X-Inertia', 'true')
         @render_method.call json: page, status: @response.status, content_type: Mime[:json]
       else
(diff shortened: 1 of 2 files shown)


bknoles:
Correct diagnosis and fix here, but I noticed that the original code overwrites the Vary header, when we should have been appending to it instead.

I pulled this branch and added a commit on top of it to handle that case in #136. Thank you @skryukov !

PR #130, opened 2024-10-04
[V2] Deferred Props

### LINKED ISSUE #129: V2: Implementing Deferred props
V2 is coming to Inertia, so let's get ready.

## Context

The concept of Deferred props is:  

> Props that are not evaluated on initial page load but are requested immediately after a page is loaded, automatically.

On the backend, they work similarly to LazyProps, but the key differences lie in how they're handled on the frontend.

## Implementation

Probably we can just create an alias to `InertiaRails.lazy` and create this new feature for grouping props

### Group

From _advanced-inertia.com_
> Inertia also lets you group deferred props. This enables you to load certain deferred groups independently from others, so that quicker queries will not have to wait until the slower ones are completed.

We have to implement a group feature too

Source:
- https://advanced-inertia.com/blog/inertia-v2
- Laravel Implementation: https://github.com/inertiajs/inertia-laravel/compare/1.x...async-requests#diff-e9c6c2799322c38d5afd5376eea239f415bc158b66fba4a7591ddaf8bb93bcc5

PedroAugustoRamalhoDuarte:
Initial PR to implement Deferred Props in rails

## TODO
- [x] Create an interface to handle new IgnoreInFirstLoadProps methods
- [x] Create `InertiaRails.defer` method
- [x] Group feature for `InertiaRails.defer`
- [x] Share the Deffered Props inside deferredProps inertia response

closes #129 

inspired by:
- https://github.com/inertiajs/inertia-laravel/compare/1.x...async-requests

### DIFF
--- a/lib/inertia_rails/ignore_first_load_prop.rb
+++ b/lib/inertia_rails/ignore_first_load_prop.rb
@@ -1,5 +1,5 @@
 module InertiaRails
-  class Lazy
+  class IgnoreFirstLoadProp
     def initialize(value = nil, &block)
       @value = value
       @block = block
--- a/lib/inertia_rails/renderer.rb
+++ b/lib/inertia_rails/renderer.rb
@@ -50,7 +50,7 @@ module InertiaRails
         if rendering_partial_component?
           key.in? partial_keys
         else
-          !prop.is_a?(InertiaRails::Lazy)
+          !prop.is_a?(InertiaRails::IgnoreFirstLoadProp)
         end
       end
 
--- a/lib/inertia_rails/renderer.rb
+++ b/lib/inertia_rails/renderer.rb
@@ -63,18 +63,32 @@ module InertiaRails
     end
 
     def page
-      {
+      default_page = {
         component: component,
         props: computed_props,
         url: @request.original_fullpath,
         version: ::InertiaRails.version,
       }
+
+      deferred_props = deferred_props_keys
+      default_page[:deferredProps] = deferred_props if deferred_props.present?
+
+      default_page
     end
 
     def deep_transform_values(hash, proc)
       return proc.call(hash) unless hash.is_a? Hash
 
       hash.transform_values {|value| deep_transform_values(value, proc)}
+    end
+
+    def deferred_props_keys
+      return if rendering_partial_component?
+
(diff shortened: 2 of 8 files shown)


buhrmi:
i will test it as soon as the 2.0 beta is out

lib/inertia_rails/ignore_first_load_prop.rb
```diff
@@ -1,5 +1,5 @@
 module InertiaRails
-  class Lazy
+  class IgnoreFirstLoadProp
```
skryukov, line 2: I don't think we should rename `Lazy`, let's create a base class `IgnoreFirstLoadProp` instead to mimic what Laravel's plugin does with the interface?

lib/inertia_rails/renderer.rb
```diff
@@ -77,6 +82,15 @@ def deep_transform_values(hash, proc)
       hash.transform_values {|value| deep_transform_values(value, proc)}
     end
 
+    def deferred_props_keys
+      return if rendering_partial_component?
+
+      @props.select { |_, prop| prop.is_a?(DeferProp) }
```
skryukov, line 88: while I really love the functional approach, personally, I would optimize this with smth like:
```ruby
  @props.each_with_object({}) do |(key, prop), result|
    (result[prop.group] ||= []) << key.to_s if prop.is_a?(DeferProp)
  end
```

skryukov (commented):
Thanks @PedroAugustoRamalhoDuarte 🙌 I left a couple of comments, otherwise it looks really good!

lib/inertia_rails/ignore_first_load_prop.rb
PedroAugustoRamalhoDuarte: I like the idea too, IgnoreFirstLoad too big, and i think optional is the new nomeclature

lib/inertia_rails/renderer.rb
PedroAugustoRamalhoDuarte: Yes much better

PedroAugustoRamalhoDuarte:
@skryukov i don't know with this PR make sense alone right now that is merged into v2

PedroAugustoRamalhoDuarte:
Closing in favor of https://github.com/inertiajs/inertia-rails/pull/132

PR #132, opened 2024-10-10
Inertia.js v2.0 compatibility

### LINKED ISSUE #129: V2: Implementing Deferred props

### LINKED ISSUE #131: V2: Encrypt and Clear history
In Inertia V2, we will have one feature to encrypt and clear the history of the browser;

## Context
The main goal of this feature is to address this issue https://github.com/inertiajs/inertia/pull/1784, that is a potential security issue with the back history go back when multiple users uses the same browser

## Implementation

Inspired by Inertia Laravel's implementation, we need to complete the following task:
- [ ] Send in the page response `clearHistory` and `encryptHistory` props
- [ ] Create a function that clear the history inside the controller
- [ ] Create a new config to encryptHistory or not
- [ ] Maybe create a middleware to always encrypt the history based on config (src/EncryptHistoryMiddleware.php). Maybe for the rails world, one `before_action` is enough

skryukov:
This PR based on work from #121 and #130 and aims to add full Inertia.js v2.0 compatibility (because I want to have a branch to play with the beta 😂).

It includes:

- Encrypt History configuration – based on #121
- `DeferredProp` – from #130
- `OptionalProp` - replacement for deprecated `LazyProp`
- `MergeProp`

Updated documentation: #148

Closes #129
Closes #131

### DIFF
--- a/lib/inertia_rails/inertia_rails.rb
+++ b/lib/inertia_rails/inertia_rails.rb
@@ -1,6 +1,8 @@
 require 'inertia_rails/base_prop'
+require 'inertia_rails/ignore_on_first_load_prop'
 require 'inertia_rails/always_prop'
 require 'inertia_rails/lazy_prop'
+require 'inertia_rails/defer_prop'
 require 'inertia_rails/configuration'
 
 module InertiaRails
@@ -21,6 +23,10 @@ module InertiaRails
 
     def always(&block)
       AlwaysProp.new(&block)
+    end
+
+    def defer(group: nil, &block)
+      DeferProp.new(group:, &block)
     end
   end
 end
--- a/lib/inertia_rails/inertia_rails.rb
+++ b/lib/inertia_rails/inertia_rails.rb
@@ -2,6 +2,7 @@ require 'inertia_rails/base_prop'
 require 'inertia_rails/ignore_on_first_load_prop'
 require 'inertia_rails/always_prop'
 require 'inertia_rails/lazy_prop'
+require 'inertia_rails/optional_prop'
 require 'inertia_rails/defer_prop'
 require 'inertia_rails/merge_prop'
 require 'inertia_rails/configuration'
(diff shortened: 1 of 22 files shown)


lib/inertia_rails/inertia_rails.rb
```diff
@@ -1,52 +1,41 @@
...
+      CONFIGURATION
+    end
 
-  def self.lazy(value = nil, &block)
-    InertiaRails::Lazy.new(value, &block)
-  end
+    def optional(value = nil, &block)
+      OptionalProp.new(value, &block)
+    end
+    alias_method :lazy, :optional
```
PedroAugustoRamalhoDuarte, line 27: I think lazy will be deprecated, we can do that with alias?

PedroAugustoRamalhoDuarte:
I have created a PR into your fork to solve `X-Inertia-Partial-Except` https://github.com/skryukov/inertia-rails/pull/1

lib/inertia_rails/inertia_rails.rb
skryukov: Added the deprecation, thanks!

Issue #115, opened 2024-04-29
Rethinking Shared Data


bknoles:
Continuing the conversation from #108 

A quick summary:

* The combination of "module level" shared data storage, and "controller level" setters for shared data has introduced tricky bugs including shared data leaking across threads and across requests.
* While we've patched these issues, it'd be great to refactor away from global shared state.
* Controller class variables have been suggested in the past.
* I'm wary of using controller class variables, since in production they are also a form of global shared state.
* @PedroAugustoRamalhoDuarte has put together #111 , which refactors inertia shared state into instance variables. These are set via `before_action` callbacks, called by `inertia_share`.

bknoles:
Ok, so here are a couple examples of failing tests:

First, #117 demonstrates that `class_attribute` will leak data in the case where `inertia_share` is called conditionally. [Here's the commit](https://github.com/inertiajs/inertia-rails/pull/117/commits/7ee4349eb1e6a9629edd55d600088cb82cdfee11) with the failing example.

The situation changes a bit with the refactor to instance variables in #116. You can still get leaky data, but it's more awkward to get there. [Here's the commit](https://github.com/inertiajs/inertia-rails/pull/116/commits/01d32aef7d40797fd9226c22530dad3cb01e6c0b) with the failing test. Because of the refactor to `before_action`s, you actually have to call the route with conditional sharing twice in order to get the error. The shared data wouldn't ever be present on the first page load in development, so I'd guess this wouldn't make it into a production application.

Both of these are contrived examples that are "incorrect" uses of Inertia. But I think it's reasonable for a developer to expect to share data on a per-action basis, instead of per-controller. Especially since other Rails controller class methods like `before_action` allow per-action configuration. I think we should support it in the gem.

That said, I don't think we need to solve that in order to approve #111. It replaces global storage with the per-request storage of instance variables. That removes the necessity for a lot of current workarounds without changing the public API at all. As the PR notes, conditional sharing can be a future improvement.

bknoles:
Here a couple (untested) sketches for API changes that could support conditional rendering. Essentially just passing options to the `before_action` within `inertia_share`.

The first one requires a breaking change. Plain data props must be defined in a `props:` argument. Everything else gets passed onto `before_action`. It's the same API as the renderer, which is nice, but breaking changes are no fun.

```ruby
module InertiaRails
  module Controller
    extend ActiveSupport::Concern

    module ClassMethods
      def inertia_share(props: nil, **options, &block)
[...]

PR #137, opened 2024-10-18
`before_action` like filters for `inertia_share`

skryukov:
This PR builds on the approach from #121 and aims to resolve the concerns raised in https://github.com/inertiajs/inertia-rails/issues/115#issuecomment-2420718117.

It introduces some complexity as we need to partially reimplement Rails callbacks to work with @ElMassimo's shared properties feature.

@bknoles,  I'd like to get your opinion on this approach. Do you think it's worth it to continue refining this PR?

Supported options:

```ruby
inertia_share normal_shared_prop: 1

inertia_share(normal_shared_prop: 1)

inertia_share do
  {conditionally_shared_show_prop: 1} if action_name == "show"
end

inertia_share only: :edit do
  {only_block_prop: 1}
end

inertia_share except: [:show, :index] do
  {except_block_prop: 1}
end

inertia_share if: -> { is_edit? } do
  {if_proc_prop: 1}
end

inertia_share unless: -> { !is_edit? } do
  {unless_proc_prop: 1}
end

inertia_share({only_prop: 1}, only: :edit)

inertia_share({if_prop: 1}, if: [:is_edit?, -> { true }])

inertia_share({unless_prop: 1}, unless: :not_edit?)

inertia_share({only_if_prop: 1}, only: :edit, if: -> { true })

inertia_share({except_if_prop: 1}, except: [:index, :show], if: -> { true })
```

Caveat:
```ruby
inertia_share normal_shared_prop: 1, only: :edit
# => raises "You must not mix shared data and [:if, :unless, :only, :except] options, pass data as a hash or a block.
inertia_share(normal_shared_prop: 1, only: :edit)
# => raises "You must not mix shared data and [:if, :unless, :only, :except] options, pass data as a hash or a block.
inertia_share({normal_shared_prop: 1, only: :edit})
# => adds only: :edit as a prop
inertia_share({normal_shared_prop: 1, only: :edit}, only: :edit)
# => adds only: :edit as a prop only for :edit action
```

### DIFF
--- a/lib/inertia_rails/controller.rb
+++ b/lib/inertia_rails/controller.rb
@@ -14,10 +14,19 @@ module InertiaRails
     end
 
     module ClassMethods
-      def inertia_share(attrs = {}, &block)
-        @inertia_share ||= []
-        @inertia_share << attrs.freeze unless attrs.empty?
-        @inertia_share << block if block
+      def inertia_share(hash = nil, **props, &block)
+        options = extract_inertia_share_options(props)
+        return push_to_inertia_share(**(hash || props), &block) if options.empty?
+
+        push_to_inertia_share do
+          next unless options[:if].all? { |filter| instance_exec(&filter) } if options[:if]
+          next unless options[:unless].none? { |filter| instance_exec(&filter)  } if options[:unless]
+
+          next hash unless block
+
+          res = instance_exec(&block)
+          hash ? hash.merge(res) : res
+        end
       end
 
       def inertia_config(**attrs)
@@ -53,6 +62,53 @@ module InertiaRails
           else
             @inertia_share || shared_data || []
           end.freeze
+        end
+      end
+
+      private
+
+      def push_to_inertia_share(**attrs, &block)
+        @inertia_share ||= []
+        @inertia_share << attrs.freeze unless attrs.empty?
+        @inertia_share << block if block
+      end
+
+      def extract_inertia_share_options(props)
+        options = props.slice(:if, :unless, :only, :except)
+
+        return options if options.empty?
+
+        if props.except(:if, :unless, :only, :except).any?
+          raise ArgumentError, "You must not mix shared data and [:if, :unless, :only, :except] options, pass data as a hash or a block."
+        end
+
+        extract_inertia_share_option(options, :only, :if)
+        extract_inertia_share_option(options, :except, :unless)
+
+        options.transform_values! do |filters|
+          Array(filters).map!(&method(:filter_to_proc))
+        end
+
+        options
+      end
+
+      def extract_inertia_share_option(options, from, to)
+        if (from_value = options.delete(from))
+          filter = AbstractController::Callbacks::ActionFilter.new([:inertia_share], from, from_value)
+          options[to] = Array(options[to]).unshift(filter)
+        end
+      end
+
+      def filter_to_proc(filter)
+        case filter
+        when Symbol
+          -> { send(filter) }
+        when Proc
+          filter
+        when AbstractController::Callbacks::ActionFilter
+          -> { filter.match?(self) }
+        else
+          raise ArgumentError, "You must pass a symbol or a proc as a filter."
         end
       end
     end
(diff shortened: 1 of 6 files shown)


bknoles:
I love the API. Especially the clear and helpful exceptions if you try write something that could lead to keyword collision.

Last night when I looked at this, I worried about this as well:

> we need to partially reimplement Rails callbacks

I.e. would it be confusing for someone new if the API to kinda sorts resembles `before_action`, but actually does something different?

However, after thinking about it, I think keys like if/only/except/unless are generic enough that they don't "belong" to callbacks. The pattern here feels very much like a _Rails_ pattern, rather than a _Rails callbacks_ pattern.

I'm in favor of adding. I would use this in existing apps.

@BrandonShar can you strawman this? Are there any other potential downsides you can think of that we should consider?

lib/inertia_rails/controller.rb
```diff
@@ -55,6 +65,53 @@ def _inertia_shared_data
...
+      def extract_inertia_share_options(props)
+        options = props.slice(:if, :unless, :only, :except)
+
+        return options if options.empty?
+
+        if props.except(:if, :unless, :only, :except).any?
+          raise ArgumentError, "You must not mix shared data and [:if, :unless, :only, :except] options, pass data as a hash or a block."
+        end
+
+        extract_inertia_share_option(options, :only, :if)
```
BrandonShar, line 86: Could we add some indication that these methods are mutating their argument? maybe just a `!` suffix (even though ruby is super vague on what that means 😄 )

spec/inertia/conditional_sharing_spec.rb
```diff
@@ -14,6 +14,20 @@
       expect(JSON.parse(response.body)['props'].deep_symbolize_keys).not_to include({
         conditionally_shared_show_prop: 1,
       })
+      get conditional_share_edit_path, headers: { 'X-Inertia' => true }
+      expect(JSON.parse(response.body)['props'].deep_symbolize_keys).to eq({
```
BrandonShar, line 18: I think this spec is solid, but could we add one more spec here that verifies that these props are **not** included when a different action on this controller is requested? I always like covering the opposite of the happy path as well since I believe this spec would also pass if I deleted all of the filtering code.

BrandonShar (commented):
I like it! I think this is a nice "papercut" style improvement that feels very in line with existing expectations for rails users. 

Just two minor comments I'd prefer be addressed before we merge.

lib/inertia_rails/controller.rb
skryukov: That's a small thing, but I will stand my ground 😂

I believe generally a `!` method will have a non-bang counterpart. Another point is that Rails itself doesn't use bang for this method: https://github.com/rails/rails/blob/v7.2.0/actionpack/lib/abstract_controller/callbacks.rb#L98-L104

spec/inertia/conditional_sharing_spec.rb
skryukov: Updated the spec to use `expect.to eq` and replaced the props names to make them more obvious.

lib/inertia_rails/controller.rb
BrandonShar: Ok, I can cede on the `!` 😄 

Maybe there's a naming thing we could change here? Your example from rails is good, but I think maybe what throws me here is that we have `extra_inertia_share_options` above which is a pure function and then a very similarly named method here that mutates its arguments and has no return. Would something like `transform_inertia_share_option` make sense?

spec/inertia/conditional_sharing_spec.rb
BrandonShar: awesome thank you; new spec looks great!

lib/inertia_rails/controller.rb
skryukov: Sounds good, updated 👍

BrandonShar (approved):
Thanks for the back and forth!

PR #138, opened 2024-10-19
Vary Header should not be controlled by the client

skryukov:
The Vary header is a way for servers to control caching mechanisms, and it is purely a server-side mechanism. The client does not control or affect the Vary header. Issue #136 introduced logic that does not conform to this: https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-semantics-19#section-12.5.5

There are several approaches we can take with the Vary header:
1. Apply the original Laravel plugin logic: force the `Vary: X-Inertia` header.
2. Set the `Vary` header if it wasn't set before, which is the default logic used in Rails: https://github.com/rails/rails/blob/b8720e7cb8c9894d142b8576c21065c92cd1907a/actionpack/lib/action_controller/metal/rendering.rb#L220-L224
3. Append `X-Inertia` if the `Vary` header was set previously, or set the `Vary: X-Inertia` header if it was not.

I believe the third option makes the most sense: it allows users to add other headers, like `Accept-Language`, to the `Vary` list, but it also prevents them from omitting `X-Inertia`.

### DIFF
--- a/lib/inertia_rails/renderer.rb
+++ b/lib/inertia_rails/renderer.rb
@@ -16,7 +16,11 @@ module InertiaRails
     end
 
     def render
-      @response.set_header('Vary', [@request.headers['Vary'], 'X-Inertia'].compact.join(', '))
+      if @response.headers["Vary"].blank?
+        @response.headers["Vary"] = 'X-Inertia'
+      else
+        @response.headers["Vary"] = "#{@response.headers["Vary"]}, X-Inertia"
+      end
       if @request.headers['X-Inertia']
         @response.set_header('X-Inertia', 'true')
         @render_method.call json: page, status: @response.status, content_type: Mime[:json]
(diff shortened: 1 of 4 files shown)


lib/inertia_rails/renderer.rb
```diff
@@ -16,7 +16,11 @@ def initialize(component, controller, request, response, render_method, props: n
     end
 
     def render
-      @response.set_header('Vary', [@request.headers['Vary'], 'X-Inertia'].compact.join(', '))
+      if @response.headers["Vary"].blank?
```
skryukov, line 19: To be honest, I prefer the original functional approach, but it generates unnecessary objects, which can affect production performance due to additional GC load.

BrandonShar (commented):
Thanks for the PR @skryukov !

To make sure I understand, is this PR purely a performance optimization? It looks to me like the specs you added will pass with the existing logic, right?

skryukov:
@BrandonShar Not quite - previously we passed the value from the **request's** `Vary` header, which seems unusual since I don't see this behavior in either the Laravel plugin or the HTTP caching RFC 🤔

BrandonShar (approved):
I read over this change like 3 times before I left that comment and still missed that it said `@request` in the original line, not `@response` 😆. Thanks for calling that out!

bknoles:
Released in 3.3.0!

PR #141, opened 2024-10-25
lookup params in json when locals not present

BenMorganMY:
This is to be used when testing the inertia response while specifying `headers: { 'X-Inertia' => 'true' }` in the request spec.


BrandonShar:
Thanks for this @BenMorganMY ! It looks like this same issue was also identified by @iatanas0v in #94 and I totally missed it! I'm going to merge that one just because it was an earlier PR and it has some spec changes in it. Thank you for bumping this issue!

BrandonShar:
Closing per merge of #94

PR #142, opened 2024-10-25
Ensure the Props Being Passed Have Been Turned Into a Hash

BenMorganMY:
We encountered a case where we would pass active model errors to the response. This allows objects such as ActiveModel::Errors which can be happily serialized to a hash to be passed directly into the props without calling `to_hash` all of the time.

If the serializer your rails app is using is not :json or :marshal, such as :message_pack, you will encounter an error when passing in an ActiveModel::Errors object since they don't respond to to_msgpack. This also fixes this case by converting the instance to a hash and then allowing our serializer to work successfully without going back to :marshal or :json.

### DIFF
--- a/lib/inertia_rails/renderer.rb
+++ b/lib/inertia_rails/renderer.rb
@@ -75,7 +75,8 @@ module InertiaRails
       deep_transform_values(
         _props,
         lambda do |prop|
-          prop.respond_to?(:call) ? controller.instance_exec(&prop) : prop
+          result = prop.respond_to?(:call) ? @controller.instance_exec(&prop) : prop
+          result.respond_to?(:to_hash) ? result.to_hash : result
         end
       )
     end


BrandonShar:
Thanks @BenMorganMY ! This makes sense to me. Would you be able to add a spec that demonstrates the behavior? 

Also, just for my own curiosity: what are you using on the frontend that is consuming the message_pack response?

BenMorganMY:
@BrandonShar will do, I just have one more bug to post 😅

BenMorganMY:
@BrandonShar where would be best to test this guy?

BrandonShar:
I think it would be great to have a context like `context 'when the serializer doesn't use something jsonable'` (don't quote me directly on that name 😄 ) in the [rendering_spec](https://github.com/inertiajs/inertia-rails/blob/e203b304f26b7b633e16dc0272e6420c13fdc0fe/spec/inertia/rendering_spec.rb). 

I would love something that would fail without your change in place.

PR #143, opened 2024-10-25
Call to_hash On inertia_errors for Session Serialization

BenMorganMY:
As one can see in the screenshot below, if `ActiveModel::Errors` is passed as the error object, it will be stored in the session. If your `cookie_serializer` is `:json`, `:marshal`, or `:hybrid` this will work fine. However, Rails now uses `:message_pack` and `ActiveModel::Errors` does not respond to `to_msgpack`. This ensures that errors being passed are turned into a hash.

Screenshot 2024-10-25 at 2 40 59 PM

### DIFF
--- a/lib/inertia_rails/controller.rb
+++ b/lib/inertia_rails/controller.rb
@@ -109,7 +109,7 @@ module InertiaRails
 
     def capture_inertia_errors(options)
       if (inertia_errors = options.dig(:inertia, :errors))
-        session[:inertia_errors] = inertia_errors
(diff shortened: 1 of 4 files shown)


bknoles:
image

If I'm reading this table correctly, you've listed all the possible options, and all should be working after this change, so looks safe to merge it in!

I'll update here once we get a release out with the change.

bknoles (approved):

bknoles:
Released in `3.4.0`!

Issue #122, opened 2024-06-26
Question: how about the frontend folder/file naming convension when using inertia-rails?


xfyuan:
First, Thanks for your hard work to introduce the great gem to integrated the inertia.js into rails! 👍 

There're a few confused questions about the frontend folder/file naming convention by using it:

- What's the best way to naming the frontend folder? snakecase follow Rails way, or other?
- What's the best way to naming the frontend file? snakecase follow Rails way, or other?

And there're another question which is met from above points: 

**could we not to call the `render` in each controller method(such as `#show, #index`) expicitly, and inertia-rails can know which frontend file/component should be used through the folder/file/class/method namecast?** 

If we can, It seems to follow the Rails way perfectly.)

arni1981:
The latter is mentioned in the docs https://github.com/inertiajs/inertia-rails?tab=readme-ov-file#rails-component-and-instance-props

```ruby
class EventsController < ApplicationController
  use_inertia_instance_props

  def index
    @events = Event.all
  end

end
```
...is the same as ...

```ruby
class EventsController < ApplicationController
  def index
    render inertia: 'events/index', props: {
      events: Event.all
    }
  end
end
```

xfyuan:
> The latter is mentioned in the docs https://github.com/inertiajs/inertia-rails?tab=readme-ov-file#rails-component-and-instance-props
> 
> ```ruby
> class EventsController < ApplicationController
>   use_inertia_instance_props
> 
>   def index
>     @events = Event.all
>   end
> 
> end
> ```
> 
> ...is the same as ...
> 
> ```ruby
> class EventsController < ApplicationController
>   def index
>     render inertia: 'events/index', props: {
>       events: Event.all
>     }
>   end
> end
> ```

Ahh yes, that's it. Thanks for your so kindly reply.)

And how about the formmer question? It seems that there're different naming style in the README examples?(such as `Event/Index` and `events/index`)

PedroAugustoRamalhoDuarte:
For now, we don't have a strong convention about that. In my opinion is better to use like `Event/Index` because is the name of the file in the frontend and they use , in React at least,  `PascalCase`.

[Inertia Rails contrib]( https://github.com/skryukov/inertia_rails-contrib) uses  `PascalCase` as well

BrandonShar:
Solved via #144 !

Issue #107, opened 2023-10-23
Proper case components will not render


matthewblott:
Like React, the convention for Svelte components is to use proper case (LikeThis) but Rails expects the default paths to be lowercase. It works (so far) if I change it to lowercase but I can see issues going forward and there should be a way to accept components named conventionally.

Below is a screenshot of what I had to do to get the `show` method to work - either use the more explicit `render` method with a named path as the first argument or change the `.svelte` file to lowercase:

![Screenshot 2023-10-23 at 17 07 16](https://github.com/inertiajs/inertia-rails/assets/1231417/f47a207e-c08c-44e4-a820-864b44302062)

BrandonShar:
Hmm.. I'm not sure we could enable both naming patterns without some heavy handed checks at runtime. 

Maybe we could enable this via a setting? Would that solve the issue here?

matthewblott:
Hi @BrandonShar thanks for the prompt response. Being able to set this somewhere like the base controller should be fine. Every major front end framework uses proper case, I'm surprised it's not come up already.

BrandonShar:
Solved via #144 !

PR #144, opened 2024-10-26
Allow Default Rendering Behavior to be Customized

BrandonShar:
Issues #107 and #122 both revolve around our "convention over configuration" not being quite the right convention. Because the convention in question is largely dictated by folder structure and component naming, I think it makes sense for us to be flexible here and allow our default renderer to be transformed.

This PR introduces a new configuration option called `component_path_resolver` that allows users to configure how inertia's default component renderer resolves the component path. For example:

```ruby
inertia_config(
   component_path_resolver: -> (path, action) do
      "Storefront/#{path.camelize}/#{action.camelize}"
   end
)
```

For a controller called `MarketplaceController` with a route called `buy_item`, this would render a component path like `Storefront/Marketplace/BuyItem` instead of our current default behavior of a path like `marketplace/buy_item`

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,19 @@ are automatically disabled for that response. Instance props are only included i
 
 Automatic component name is also opt in, you must set the [`default_render`](#default_render) config value to `true`. Otherwise, you can simply `render inertia: true` for the same behavior explicitly.
 
+If the default component naming doesn't match your convention, you can define a method to resolve it however you like via the `component_path_resolver` config value. The value of this should be callable and will receive the path and action and should return a string component path.
+
+```ruby
+inertia_config(
+  component_path_resolver: ->(path, action) do
+    "Storefront/#{path.camelize}/#{action.camelize}"
+  end
+)
+
+```
+
+
+
 ### Layout 
 
 Inertia layouts use the rails layout convention and can be set or changed in the same way.
--- a/README.md
+++ b/README.md
@@ -77,6 +77,14 @@ are automatically disabled for that response. Instance props are only included i
 
 Automatic component name is also opt in, you must set the [`default_render`](#default_render) config value to `true`. Otherwise, you can simply `render inertia: true` for the same behavior explicitly.
 
+If the automatic component name doesn't match your convention, you can define a method to transform it. This method is called `inertia_render_transformer` and it will receive your path and action and should return a string component path.
+
+```ruby
+def inertia_render_transformer(path, action)
+   "#{path.camelize}/#{action.camelize}"
+end
+```
+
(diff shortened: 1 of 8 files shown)


PedroAugustoRamalhoDuarte (approved):
Good new configuration, thanks for that!! I prefer it to be a config to stay uniform with the other ones, but the override way is fine too.

README.md
```diff
@@ -77,6 +77,14 @@ are automatically disabled for that response. Instance props are only included i
 
 Automatic component name is also opt in, you must set the [`default_render`](#default_render) config value to `true`. Otherwise, you can simply `render inertia: true` for the same behavior explicitly.
 
+If the automatic component name doesn't match your convention, you can define a method to transform it. This method is called `inertia_render_transformer` and it will receive your path and action and should return a string component path.
+
+```ruby
+def inertia_render_transformer(path, action)
```
skryukov, line 83: The name is a bit vague, could it be `inertia_page_resolver` or smth like that? I'm not good at naming, so it's mostly an open question 😅

skryukov (commented):
Love it! 🚀

I'd also prefer this to be a configuration option (will try to hack smth to see if we can make the implementation less verbose)—the reason is uniformity and discoverability as a result.

README.md
BrandonShar: Ooh, I really like resolver. How about `component_path_resolver`? Since it's moved back into the config, we don't need the inertia prefix anymore.

BrandonShar:
Good feedback! I moved it back to a configuration method and was able to get around the callable behavior by just teaching the config to not redefine already existing methods.

README.md
skryukov: Yup, sounds great :shipit:

lib/inertia_rails/configuration.rb
```diff
@@ -9,6 +9,9 @@ class Configuration
       # Overrides Rails default rendering behavior to render using Inertia by default.
       default_render: false,
 
+      # Allows the user to hook into the default rendering behavior and change it to fit their needs
+      component_path_resolver: ->(path, action) { "#{path}/#{action}" },
```
skryukov, line 13: Another quick idea (srr 😅): should we use props instead of args — that way we can make the interface less fragile: `->(path:, action:) { "#{path}/#{action}" }`. Say, in case we decide to pass controller in the future (if someone comes with a good case for that).

BrandonShar: Good call! Could save us from having to bump a major version in the future.

bknoles:
Released in 3.3.0!

PR #151, opened 2024-11-01
Use Algolia for search

skryukov:
This PR adds Algolia DocSearch API keys.

### DIFF
--- a/docs/.vitepress/config.mts
+++ b/docs/.vitepress/config.mts
@@ -138,7 +138,12 @@ export default defineConfig({
     },
 
     search: {
-      provider: 'local',
+      provider: 'algolia',
+      options: {
+        appId: 'BWKGTG68ZO',
+        apiKey: '06bc959e3f6ab7eb186cd27653408b04',
+        indexName: 'inertia-rails',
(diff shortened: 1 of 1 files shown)


docs/.vitepress/config.mts
```diff
@@ -138,7 +138,12 @@ export default defineConfig({
     },
 
     search: {
-      provider: 'local',
+      provider: 'algolia',
+      options: {
+        appId: 'BWKGTG68ZO',
+        apiKey: '06bc959e3f6ab7eb186cd27653408b04',
```
bknoles, line 144: is there risk to having this in the repo? i can throw (a regenerated) key in a Github Actions secret if we don't want this public.

skryukov: It's ok. These are client-side keys, so they are rendered in the html anyways, like google analytics: https://docsearch.algolia.com/docs/docsearch-v3#implementation

bknoles: ah right, i should have know 👍

bknoles (approved):

PR #152, opened 2024-11-02
Add support for `always` and `except`

skryukov:
This PR "upstreams" some features from #132 which are related to Inertia.js 1.x, it includes:

- New `InertiaRails.always` prop from inertia.js 1.3: https://inertiajs.com/partial-reloads#lazy-data-evaluation
- Support for the `except` option from inertia.js 1.3: https://inertiajs.com/partial-reloads#except-certain-props
- Lazy props bug fix: this PR also fixes `InertiaRails.lazy(false)` and `InertiaRails.lazy(nil)` raising errors
- Simplification for lazy props: we don't wrap simple values into procs anymore
- Performance fix for merging props: we don't symbolyze keys deeply anymore when deep merge config is turned off
- `request.inertia_partial?` update: with introduction of the `X-Inertia-Partial-Except` header, `X-Inertia-Partial-Data` is now can be omitted when `router.visit(path, {except: 'foo'})` is called.

PR is intended to be reviewed commit by commit.

### DIFF
--- a/lib/inertia_rails/base_prop.rb
+++ b/lib/inertia_rails/base_prop.rb
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 
 module InertiaRails
-  class Lazy
+  # Base class for all props.
+  class BaseProp
     def initialize(value = nil, &block)
       raise ArgumentError, 'You must provide either a value or a block, not both' if value && block
 
--- a/lib/inertia_rails/base_prop.rb
+++ b/lib/inertia_rails/base_prop.rb
@@ -3,19 +3,12 @@
 module InertiaRails
   # Base class for all props.
   class BaseProp
-    def initialize(value = nil, &block)
-      raise ArgumentError, 'You must provide either a value or a block, not both' if value && block
-
-      @value = value
+    def initialize(&block)
       @block = block
     end
 
     def call(controller)
-      value.respond_to?(:call) ? controller.instance_exec(&value) : value
-    end
-
-    def value
-      @value.nil? ? @block : @value
+      controller.instance_exec(&@block)
     end
   end
 end
--- a/lib/inertia_rails/renderer.rb
+++ b/lib/inertia_rails/renderer.rb
@@ -76,12 +76,16 @@ module InertiaRails
         end
       end
 
-      deep_transform_values(
-        _props,
-        lambda do |prop|
-          prop.respond_to?(:call) ? controller.instance_exec(&prop) : prop
+      deep_transform_values _props do |prop|
+        case prop
+        when Lazy
+          prop.call(controller)
+        when Proc
+          controller.instance_exec(&prop)
+        else
+          prop
         end
-      )
+      end
     end
 
     def page
@@ -93,10 +97,10 @@ module InertiaRails
       }
     end
 
-    def deep_transform_values(hash, proc)
-      return proc.call(hash) unless hash.is_a? Hash
+    def deep_transform_values(hash, &block)
+      return block.call(hash) unless hash.is_a? Hash
 
-      hash.transform_values {|value| deep_transform_values(value, proc)}
+      hash.transform_values {|value| deep_transform_values(value, &block)}
     end
 
     def partial_keys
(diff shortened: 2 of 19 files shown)


lib/inertia_rails/renderer.rb
```diff
@@ -64,24 +66,34 @@ def shared_data
...
-          !prop.is_a?(InertiaRails::Lazy)
+          !prop.is_a?(LazyProp)
         end
       end
 
-      deep_transform_values(
-        _props,
-        lambda do |prop|
-          prop.respond_to?(:call) ? controller.instance_exec(&prop) : prop
+      drop_partial_except_keys(_props) if rendering_partial_component?
```
bknoles, line 85: A possible nitpick, but could we return a new hash from this method instead of mutating `_props` directly?

lib/inertia_rails/renderer.rb
```diff
@@ -93,16 +105,30 @@ def page
...
+      return block.call(hash) unless hash.is_a? Hash
 
-      hash.transform_values {|value| deep_transform_values(value, proc)}
+      hash.transform_values {|value| deep_transform_values(value, &block)}
+    end
+
+    def drop_partial_except_keys(hash)
+      partial_except_keys.each do |key|
+        parts = key.to_s.split('.').map(&:to_sym)
+        *initial_keys, last_key = parts
```
bknoles, line 117: this support dot notation, yea? so like `except: ['users.some_association.expensive_thing_to_compute']`?

do the other adapters support that? I don't see it documented anywhere.

either way, if we support it for `:except`, should we also support it for `:only`? (genuine question)

lib/inertia_rails/base_prop.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module InertiaRails
+  # Base class for all props.
+  class BaseProp
```
bknoles, line 5: I like this. Half-baked thought: I wonder if we could move more rendering logic into the `Prop` classes. The `computed_props` method is doing more and more things in a sort of procedural style. Maybe it could iterate over the props one time and each instance would know what to do with itself.

(Not suggesting changes, more just thinking out loud)

lib/inertia_rails/renderer.rb
skryukov: > do the other adapters support that? I don't see it documented anywhere.
yup, Laravel adapter supports that

> if we support it for :except, should we also support it for :only?
Laravel plugin does not 😄 To be honest the whole `except` + `only` thing is poorly documented, we probably should help the core team make it happen.

lib/inertia_rails/renderer.rb
skryukov: I'm trying not to create new objects when possible since we're working on the framework level and our implementation can affect app performance. Maybe I'm to protective here though. Need to come up with a good benchmarking example to test such things.

lib/inertia_rails/renderer.rb
skryukov: Actually they reverted back dot notation for `only`: https://github.com/inertiajs/inertia-laravel/commit/4ce06f6fafe9405e5c58b38638de33e0240c53e8

bknoles: Oh nice find!

bknoles: Ok, i dug into this a bit. TL;DR: I don't think those issues will affect our implementation.

Best I can tell (caveat: using my brain's PHP interpreter) is that the original implementation of Laravel's `share` shallow merged all data. However, dot notation allowed you to deep merge dot-notated shared data. Not sure whether that was intentional or a happy accident since I don't think the feature was ever documented. When dot notation was added to `only` and `except`, it changed the execution order and changed the behavior such that _all_ shared data was a shallow merge. So, people relying on the deep merge behavior had breaking changes. (There was another reported but with lazy props that I didn't fully diagnose... looks to be something related to array manipulation quirks in PHP).

InertiaRails 1) has explicit patterns for deep merging shared data and 2) doesn't support dot notation in shared data anyway, so we should be totally fine.

I don't see a reason why we couldn't support dot notation for `only` as well. Do you think it's worth adding? Seems like it'd be nice if some prop had multiple expensive child props and you only wanted to grab one of them.

skryukov: Yup, let's try adding support for ’only’ in a separate PR 👍

bknoles (approved):

PR #158, opened 2024-11-11
Upstream generators

skryukov:
This PR upstreams generators from https://github.com/skryukov/inertia_rails-contrib


jacobmichels:
This is amazing, I'd really like to use the typescript scaffolds you've created. Do you need any assistance in getting this PR ready for review?

skryukov:
@jacobmichels thanks for the ping! I hope to finish this PR in couple days, my current goal is to implement simple smoke test that will generate new Rails app, install inertia, generate some scaffolds and run system tests.

lib/generators/inertia/install/js_package_manager.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+module Inertia
+  module Generators
+    class JSPackageManager
+      def self.package_managers
+        %w[npm yarn bun pnpm]
```
bknoles, line 7: Gonna point to this file next time I need an example of why the JS ecosystem feels like the wild west compared to Ruby 😅

spec/generators/install/install_generator_spec.rb
```diff
@@ -0,0 +1,314 @@
+# frozen_string_literal: true
+
+require_relative '../../../lib/generators/inertia/install/install_generator'
+require 'generator_spec'
+
+RSpec.describe Inertia::Generators::InstallGenerator, type: :generator do
```
bknoles, line 6: this is really cool!

bknoles (approved):
so excited to have this upstreamed! thanks @skryukov

PR #163, opened 2024-11-14
Support dot notation for :only keys in partial reloads

### LINKED ISSUE #157: Add support for dot notation in :only partial reloads
#152 added support for the `:except` prop, including dot notation:

```
except: ['some.nested.path']
```

Would be nice to have this available for the `:only` headers as well!

bknoles:
Resolves #157 

I also refactored the rendering logic a bit, to:

1. Only traverse the props hash a single time after it's merged with shared props
2. Put the prop filtering logic in a single place

Note: the renderer diff isn't very helpful; better to view the entire file to see how it looks now.

TODO:

- [x] add some more specs to test the interaction of all these crazy prop filtering types 😅
- [x] update docs to explain dot notation

### DIFF
--- a/lib/inertia_rails/renderer.rb
+++ b/lib/inertia_rails/renderer.rb
@@ -105,7 +105,7 @@ module InertiaRails
       props.reduce({}) do |transformed_props, (key, prop)|
         current_path = parent_path + [key]
 
-        if prop.is_a?(Hash)
+        if prop.is_a?(Hash) && prop.any?
           nested = deep_transform_props(prop, current_path, &block)
           transformed_props.merge!(key => nested) unless nested.empty?
         else
--- a/lib/inertia_rails/renderer.rb
+++ b/lib/inertia_rails/renderer.rb
@@ -6,6 +6,9 @@ require_relative "inertia_rails"
 
 module InertiaRails
   class Renderer
+    KEEP_PROP = :keep
+    DONT_KEEP_PROP = :dont_keep
+
     attr_reader(
       :component,
       :configuration,
@@ -77,7 +80,7 @@ module InertiaRails
       _props = merge_props(shared_data, props)
 
       deep_transform_props _props do |prop, path|
-        next [:dont_keep] unless keep_prop?(prop, path)
+        next [DONT_KEEP_PROP] unless keep_prop?(prop, path)
 
         transformed_prop = case prop
         when BaseProp
@@ -88,7 +91,7 @@ module InertiaRails
           prop
         end
 
-        [:keep, transformed_prop]
+        [KEEP_PROP, transformed_prop]
       end
     end
 
@@ -110,7 +113,7 @@ module InertiaRails
           transformed_props.merge!(key => nested) unless nested.empty?
         else
           action, transformed_prop = block.call(prop, current_path)
-          transformed_props.merge!(key => transformed_prop) if action == :keep
+          transformed_props.merge!(key => transformed_prop) if action == KEEP_PROP
         end
 
         transformed_props
--- a/lib/inertia_rails/renderer.rb
+++ b/lib/inertia_rails/renderer.rb
@@ -74,18 +74,12 @@ module InertiaRails
     end
 
     def computed_props
-      _props = merge_props(shared_data, props).select do |key, prop|
-        if rendering_partial_component?
-          partial_keys.none? || key.in?(partial_keys) || prop.is_a?(AlwaysProp)
-        else
-          !prop.is_a?(LazyProp)
-        end
-      end
+      _props = merge_props(shared_data, props)
 
-      drop_partial_except_keys(_props) if rendering_partial_component?
+      deep_transform_props _props do |prop, path|
+        next [:dont_keep] unless keep_prop?(prop, path)
 
-      deep_transform_values _props do |prop|
-        case prop
+        transformed_prop = case prop
         when BaseProp
           prop.call(controller)
         when Proc
@@ -93,6 +87,8 @@ module InertiaRails
         else
           prop
         end
+
+        [:keep, transformed_prop]
       end
     end
 
@@ -103,6 +99,22 @@ module InertiaRails
         url: @request.original_fullpath,
         version: configuration.version,
       }
+    end
+
+    def deep_transform_props(props, parent_path = '', &block)
+      props.reduce({}) do |transformed_props, (key, prop)|
+        current_path = [parent_path, key].reject(&:empty?).join('.')
+
+        if prop.is_a?(Hash)
+          nested = deep_transform_props(prop, current_path, &block)
+          transformed_props.merge!(key => nested) unless nested.empty?
+        else
+          action, transformed_prop = block.call(prop, current_path)
+          transformed_props.merge!(key => transformed_prop) if action == :keep
+        end
+
+        transformed_props
+      end
     end
 
     def deep_transform_values(hash, &block)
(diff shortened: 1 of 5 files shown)


lib/inertia_rails/renderer.rb
```diff
@@ -74,25 +74,21 @@ def merge_props(shared_data, props)
...
-          partial_keys.none? || key.in?(partial_keys) || prop.is_a?(AlwaysProp)
-        else
-          !prop.is_a?(LazyProp)
-        end
-      end
+      _props = merge_props(shared_data, props)
 
-      drop_partial_except_keys(_props) if rendering_partial_component?
+      deep_transform_props _props do |prop, path|
+        next [:dont_keep] unless keep_prop?(prop, path)
```
bknoles, line 80: Why return this array from the block now? Well, the block has two jobs 1) to filter out props we don't need and 2) transform the keepers into their final format. We want track those separately so that the transformation can return things that are nil/false-y.

(This block will also be a nice place to rewrite the logic from #154 )

lib/inertia_rails/renderer.rb
```diff
@@ -105,19 +101,19 @@ def page
       }
     end
 
-    def deep_transform_values(hash, &block)
-      return block.call(hash) unless hash.is_a? Hash
-
-      hash.transform_values {|value| deep_transform_values(value, &block)}
-    end
+    def deep_transform_props(props, parent_path = '', &block)
```
bknoles, line 104: Similar logic as the previous method, but we're now also dealing with the results of the filtering within the hash we're building up and conditionally setting keys.

lib/inertia_rails/renderer.rb
```diff
@@ -138,5 +134,35 @@ def resolve_component(component)
 
       configuration.component_path_resolver(path: controller.controller_path, action: controller.action_name)
     end
+
+    def keep_prop?(prop, path)
```
bknoles, line 138: all the prop filtering logic now lives in one concise place. i debated pulling this out into a class, but I don't think the cleanliness is worth the extra indirection.

lib/inertia_rails/renderer.rb
```diff
@@ -138,5 +134,35 @@ def resolve_component(component)
...
+      return true if prop.is_a?(AlwaysProp)
+
+      if rendering_partial_component?
+        path_with_prefixes = path_prefixes(path)
```
bknoles, line 142: Maybe worth nothing that if we're in a partial reload, we'll end up calculating this array for every single prop key. That should still be way faster than fully evaluating all of those props, though.

lib/inertia_rails/renderer.rb
```diff
@@ -138,5 +134,35 @@ def resolve_component(component)
...
+        parts[0..i].join('.')
+      end
+    end
+
+    def excluded_by_only_partial_keys?(path_with_prefixes)
+      partial_keys.present? && (path_with_prefixes & partial_keys.map(&:to_s)).empty?
+    end
+
+    def excluded_by_except_partial_keys?(path_with_prefixes)
+      partial_except_keys.present? && (path_with_prefixes & partial_except_keys.map(&:to_s)).any?
```
skryukov, line 165: I think we can replace `.filter_map(&:to_sym)` in the `partial_except_keys` and then stop calling `.map(&:to_s)` here

And same for `partial_keys`

```suggestion
      partial_except_keys.present? && (path_with_prefixes & partial_except_keys).any?
```

lib/inertia_rails/renderer.rb
```diff
@@ -105,19 +101,19 @@ def page
...
+    def deep_transform_props(props, parent_path = '', &block)
+      props.reduce({}) do |transformed_props, (key, prop)|
+        current_path = [parent_path, key].reject(&:empty?).join('.')
 
-    def drop_partial_except_keys(hash)
-      partial_except_keys.each do |key|
-        parts = key.to_s.split('.').map(&:to_sym)
-        *initial_keys, last_key = parts
-        current = initial_keys.any? ? hash.dig(*initial_keys) : hash
+        if prop.is_a?(Hash)
```
skryukov, line 108: I noticed that `'what_about_empty_hash' => {},` example will fail, here's a fix:

```suggestion
        if prop.is_a?(Hash) && prop.any?
```

lib/inertia_rails/renderer.rb
```diff
@@ -105,19 +101,19 @@ def page
...
     end
 
-    def deep_transform_values(hash, &block)
-      return block.call(hash) unless hash.is_a? Hash
-
-      hash.transform_values {|value| deep_transform_values(value, &block)}
-    end
+    def deep_transform_props(props, parent_path = '', &block)
+      props.reduce({}) do |transformed_props, (key, prop)|
+        current_path = [parent_path, key].reject(&:empty?).join('.')
```
skryukov, line 106: we can simplify this by sticking with arrays:

```suggestion
        current_path = path + [key]
```

skryukov (commented):
Love this, looks really clean!

bknoles:
Documentation here:

[https://dot-notation-for-only-props.inertia-rails.pages.dev/guide/partial-reloads#dot-notation](https://dot-notation-for-only-props.inertia-rails.pages.dev/guide/partial-reloads#dot-notation)

lib/inertia_rails/renderer.rb
BrandonShar, line 80: nitpick: could we move these symbol values to constants since they're acting as an enum?

BrandonShar (approved):
Looks good! only one small comment

lib/inertia_rails/renderer.rb
bknoles: sure! as an aside, `:dont_keep` doesn't do anything at all... it's just there to reveal intention

chasegiunta:
@bknoles I spent an absurd amount of time troubleshooting why `only: ['nested.prop']` was wiping out the entire top-level prop instead of just reloading the nested one, before realizing this isn't a core Inertia feature, so the frontend has no idea the response would need to be merged in.

Is it possible to add this caveat to the rails documentation?

chasegiunta:
@bknoles sure. I mean, easiest example would be I have a `current_user` shared prop. It looks something like this:

```
current_user: {
 first_name,
 last_name,
 email,
 organizations: { ... },
 preferences: { ... },
 slug
}
```

Writing a profile form, updating just the user's first & last name (no need to reload all the orgs, preferences, etc again). So we `.post()` and use `only: ['current_user.first_name', 'current_user.last_name']`. Doing this will force a reload of the entire `current_user` object prop and so the only thing returned from Inertia is 

```
current_user: {
  first_name,
  last_name
}
```

Well, that's not ideal. Now if I have email, orgs, etc displayed elsewhere on the page (not in forms, just displayed in the UI), all that data is lost.

Why it doesn't work makes sense: it's not a Inertia core feature. The Inertia frontend library would have to be aware of which properties being returned in the response are meant to be merged into frontend props. It doesn't currently do that.

bknoles:
Oh I see. Yea, partial data reloads overwrite props rather than merging them. The use case for dot notation :only keys in the [Laravel adapter's test suite](https://github.com/inertiajs/inertia-laravel/blob/2.x/tests/ResponseTest.php#L691) is instructive: the :only data really just serves to _filter out_ a `token` property that isn't necessary at all on reloads. That's different than not wanting to reload something like `organizations` while still wanting them around in the UI.

Have you looked at the [merging props](https://inertia-rails.dev/guide/merging-props) section of the docs? I think it might be what you're after? Something like:

```ruby
def create
  render 'Some/Page', props: {
    # :only keys should filter this down to a tiny response; performance benefits depends on how you render the JSON
    current_user: InertiaRails.merge(current_user)
  }
end
```

(caveat: i haven't built merge props into one of our apps so there may be a nuance my cursory review isn't capturing)

chasegiunta:
@bknoles I tried playing with it, yeah, but couldn't quite get it. I think the syntax would actually be `current_user: InertiaRails.merge { current_user }`, right? (based on looking at the docs, not a ruby dev so forgive me), but we're using a serializer for the current_user anyway. Using `only: ['current_user.first_name']` there would just omit `current_user` entirely in the response

bknoles:
Ah yea, merging won't work either now that I think it through. We actually call this out in the docs:

image

The purpose of a partial reload is to prevent unnecessary computations by filtering out props _before_ any callable props are evaluated. InertiaRails.merge() is a callable. From the perspective of the filtering algorithm, there are zero users with a "name" property in this code:

```ruby
render inertia: 'SomePage', props: {
  users: -> { User.where(some_conditions) }
 }
```

because the filtering happens before `User.where(some_conditions)` is evaluated.

Maybe you could use a query parameter to render the user differently within a merge prop (I haven't tested this):

```ruby
render inertia: 'SomePage', props: {
  current_user: InertiaRails.deep_merge{
    params[:give_me_the_slim_user_please] ? current_user.as_json(only: [:first_name, :last_name]) : current_user
   }
}
```

then in the JS

```javascript
router.get(currentPagePath, { give_me_the_slim_user_please: 1 })
```

Basically, grabbing a subset of an evaluated prop and then merging it back into the Page isn't a workflow that partial reloading is capable of doing right now... so some creativity is required.

I will say that we're talking about this and would like to support complex fetching/merging more cleanly. But nothing ready as of today.

PR #164, opened 2024-11-16
Avoid monkey patching

adrianpacala:
Refactors the patches directory to leverage Ruby's module prepend mechanism instead of direct `class_eval` calls. This will introduce dedicated modules under the `InertiaRails` namespace that are systematically prepended to their target classes.

This architectural change offers several benefits:

- Improves debugging capabilities through clearer stack traces
- Better encapsulation of Rails core modifications
- More maintainable and predictable method lookup chain

The new approach provides a cleaner separation of concerns while maintaining the same functional behavior of the original implementation.

### DIFF
--- a/lib/patches/better_errors.rb
+++ b/lib/patches/better_errors.rb
@@ -4,14 +4,16 @@
 # https://github.com/BetterErrors/better_errors/blob/v2.5.1/lib/better_errors/middleware.rb
 #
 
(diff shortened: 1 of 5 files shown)


bknoles (approved):

bknoles:
looks good! Thanks @adrianpacala

PR #165, opened 2024-11-26
Unoptimized partial reload warnings

bknoles:
I revisited this after:

1. Refactoring the renderer a bit in #163 
2. Perusing the Rails source for the behavior on unpermitted parameters (which is a analogous to this feature IMO)

The refactored renderer was much easier to add this feature to, and I liked Rails' use of ActiveSupport::Notifications for logging/raising errors. So here we are!

### DIFF
--- a/lib/inertia_rails/engine.rb
+++ b/lib/inertia_rails/engine.rb
@@ -12,5 +12,20 @@ module InertiaRails
         include ::InertiaRails::Controller
       end
     end
+
+    initializer "inertia_rails.subscribe_to_notifications" do
+      if Rails.env.development? || Rails.env.test?
+        ActiveSupport::Notifications.subscribe('inertia_rails.unoptimized_partial_render') do |*args|
+          event = ActiveSupport::Notifications::Event.new(*args)
+
+          message =
+            "InertiaRails: The \"#{event.payload[:paths].join(', ')}\" " \
+          "prop(s) were excluded in a partial reload but still evaluated because they are defined as values. " \
+          "Consider wrapping them in something callable like a lambda."
+
+          Rails.logger.debug(message)
+        end
+      end
+    end
   end
 end
--- a/lib/inertia_rails/renderer.rb
+++ b/lib/inertia_rails/renderer.rb
@@ -30,21 +30,28 @@ module InertiaRails
     end
 
     def render
+      set_vary_header
+
+      return render_inertia_response if @request.headers['X-Inertia']
+      return render_ssr if configuration.ssr_enabled rescue nil
+
+      render_full_page
+    end
+
+    private
+
+    def set_vary_header
       if @response.headers["Vary"].blank?
         @response.headers["Vary"] = 'X-Inertia'
       else
         @response.headers["Vary"] = "#{@response.headers["Vary"]}, X-Inertia"
       end
-      if @request.headers['X-Inertia']
-        @response.set_header('X-Inertia', 'true')
-        @render_method.call json: page, status: @response.status, content_type: Mime[:json]
-      else
-        return render_ssr if configuration.ssr_enabled rescue nil
-        @render_method.call template: 'inertia', layout: layout, locals: view_data.merge(page: page)
-      end
     end
 
-    private
+    def render_inertia_response
+      @response.set_header('X-Inertia', 'true')
+      @render_method.call json: page, status: @response.status, content_type: Mime[:json]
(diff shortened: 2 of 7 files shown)


lib/inertia_rails/engine.rb
```diff
@@ -12,5 +12,20 @@ class Engine < ::Rails::Engine
         include ::InertiaRails::Controller
       end
     end
+
+    initializer "inertia_rails.subscribe_to_notifications" do
+      if Rails.env.development? || Rails.env.test?
```
bknoles, line 17: I want the default experience to add warnings to the logs in development / test _only_.

lib/inertia_rails/renderer.rb
```diff
@@ -30,21 +27,30 @@ def initialize(component, controller, request, response, render_method, props: n
...
 
     def render
+      set_vary_header
+
+      validate_partial_reload_optimization if rendering_partial_component?
+
+      return render_inertia_response if @request.headers['X-Inertia']
+      return render_ssr if configuration.ssr_enabled rescue nil
+
+      render_full_page
```
bknoles, line 37: Moved these into methods purely for readability. IMO this method should be super readable since it spells out the core functionality

lib/inertia_rails/renderer.rb
```diff
@@ -68,58 +78,28 @@ def shared_data
     #
     # Functionally, this permits using either string or symbol keys in the controller. Since the results
     # is cast to json, we should treat string/symbol keys as identical.
-    def merge_props(shared_data, props)
-      if @deep_merge
-        shared_data.deep_symbolize_keys.deep_merge!(props.deep_symbolize_keys)
+    def merged_props
+      @merged_props ||= if @deep_merge
```
bknoles, line 82: memoizing just to future proof. it's only called once, but can't hurt

lib/inertia_rails/renderer.rb
```diff
@@ -166,5 +146,72 @@ def excluded_by_only_partial_keys?(path_with_prefixes)
     def excluded_by_except_partial_keys?(path_with_prefixes)
       partial_except_keys.present? && (path_with_prefixes & partial_except_keys).any?
     end
+
+    def validate_partial_reload_optimization
```
bknoles, line 150: don't love the long method name... i'm open to suggestions!

lib/inertia_rails/renderer.rb
```diff
@@ -166,5 +146,72 @@ def excluded_by_only_partial_keys?(path_with_prefixes)
...
+      if prop_transformer.unoptimized_prop_paths.any?
+        case configuration.action_on_unoptimized_partial_reload
+        when :log
+          ActiveSupport::Notifications.instrument(
+            'inertia_rails.unoptimized_partial_render',
+            paths: prop_transformer.unoptimized_prop_paths,
+            controller: controller,
+            action: controller.action_name,
```
bknoles, line 158: @skryukov you mentioned wanting this feature to accept a callable... with this anyone can register a subscriber with that sort of functionality. any other options you'd like to see passed to the subscriber context?

lib/inertia_rails/renderer.rb
```diff
@@ -166,5 +146,72 @@ def excluded_by_only_partial_keys?(path_with_prefixes)
...
+          )
+        when :raise
+          raise InertiaRails::UnoptimizedPartialReload.new(prop_transformer.unoptimized_prop_paths)
+        end
+      end
+    end
+
+    class PropTransformer
```
bknoles, line 166: I moved the transformation logic into a nested subclass so 1) all the behavior is encapsulated in a single place and 2) I had a place to store unoptimized_prop_paths while traversing the props hash. The transformation code is the same, and it was super easy to drop in the `track_unoptimized_prop` call.

The caller (above) uses a `.select_transformed` method with a block, so the caller only has to worry about filtering now (instead of both filtering and transforming). We also get to remove the admittedly-clunky keep/dont keep enum!

I considered pulling this out of the renderer class, but I don't really see a need for it. It's tested by our request specs and never going to be used outside that context.

spec/inertia/configuration_spec.rb
```diff
@@ -153,11 +153,60 @@
       end
     end
   end
+
+  describe '.action_on_unoptimized_partial_reload' do
```
bknoles, line 157: Much cleaner specs than in #154  :)

PR #167, opened 2024-11-28
Add new recipe how to deploy an Inertia Rails app with Kamal and SSR support

brodienguyen:
This PR adds a new recipe to the Cookbook list, providing detailed steps on how to deploy an Inertia Rails application with Kamal, configured for SSR support.

BrodieMFL 2024-12-04 at 21 24 58

> Special thanks to @skryukov for taking your time to guide me through this journey.


docs/cookbook/deploy-with-kamal.md
```diff
@@ -0,0 +1,222 @@
+# Deploy with `Kamal`
+
+Rails 8 will ship with [Kamal](https://kamal-deploy.org/) preconfigured as the default deployment tool.
```
skryukov, line 3: This is true when you generate a new app with `rails new -j esbuild` or other builders. By default (with importmap), you'll be missing Node.js-related dependencies.

mpressen: Is this comment related to this line though ?
'Cause I'm pretty sure that Kamal is the default deployment tool for rails 8, even without `-j esbuild` flag

brodienguyen: @skryukov I did update my PR to highlight the difference of example Dockerfile compared to the fresh default Rails 8. The preview of the new docs have been updated in the PR description as well. Could you help me to re-review it, please?

docs/cookbook/deploy-with-kamal.md
```diff
@@ -0,0 +1,233 @@
...
+the correct SSR server URL in the adapter. By default, the adapter points to `http://localhost:13714`,
+but this must align with the **_VITE_RUBY_HOST_** value defined in your `deploy.yml` when we deploy it to production.
+
+Update the configuration in `config/initializers/inertia_rails.rb` to dynamically construct the
+SSR URL using Vite Ruby's protocol, host, and port settings:
+
+```ruby
+InertiaRails.configure do |config|
+  config.ssr_enabled = ViteRuby.config.ssr_build_enabled
+  config.ssr_url = "#{ViteRuby.config.protocol}://#{ViteRuby.config.host}:13714" # [!code ++]
```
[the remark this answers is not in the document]
ElMassimo: For clarity, the default value of `ssr_url` was added in:

- #74

`inertia_rails` does not depend on `vite_ruby`, nor is `vite_ruby` so commonly used that it would make sense to auto-detect it and use it as a default.

In addition, I would suggest to avoid using `ViteRuby.config`, as it refers to the configuration for the _Vite **Dev** Server_, and __not__ the SSR server.

If it works, it's by coincidence.

For example, a common setup consists in running the SSR server in the same box and using plain HTTP (without SSL), while in many cases it's desirable to run the Vite dev server over SSL so that HTTP2 can be used and connections are multiplexed (loading unbundled files faster in development).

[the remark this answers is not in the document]
skryukov: Yeah, I guess this part is a leftover from when I tried to understand what `bin/vite ssr` does (which is basically just a shortcut) 😅

I think it's better to use something like this instead: `ssr_url: ENV.fetch('INERTIA_SSR_URL', 'http://localhost:13714')` and set `INERTIA_SSR_URL` in the Kamal deployment config (any suggestions for a better name?).

mpressen, line 219: Is adding this line necessary ? Without it the default seems to work just fine.

bknoles:
Question for the room... why do you want to use SSR?

(genuine question, no right/wrong answers, just trying to understand how people use Inertia in the real world)

mpressen:
> Question for the room... why do you want to use SSR?
> 
> (genuine question, no right/wrong answers, just trying to understand how people use Inertia in the real world)

The app I'm working on needs to be SEO friendly. I believe SSR is much better than CSR on that matter.

brodienguyen:
> Question for the room... why do you want to use SSR?
> 
> (genuine question, no right/wrong answers, just trying to understand how people use Inertia in the real world)

Ignore the benefit of SEO, in my case, purely because I would like to propose Inertia to the team where we are building and maintain a very legacy B2B software product. There are some requirements about backward compatibility on older devices or browsers with limited Javascript support with client so we have to play a bit safer on this transition.

Also since there is a lot of CRUD screen where we barely working with form and doesn't change that frequently, with SSR we can cache it a lot longer to save some computation and loading time for our user too. Without JS, at minimum they can still submit the form and get the job done rather than raising a bunch of support ticket to us (I am still not 100% sure if it is achievable with how InertiaJS form work).

brodienguyen:
> Question for the room... why do you want to use SSR?
> 
> (genuine question, no right/wrong answers, just trying to understand how people use Inertia in the real world)

Ignore the benefit of SEO, in my case, purely because I would like to propose Inertia to the team where we are building and maintain a very legacy B2B software product. There are some requirements about backward compatibility on older devices or browsers with limited Javascript support in the existing contract so we have to play a bit safer on this transition.

Also since there are a lot of CRUD screens where their pages are built purely with form and don't change that frequently, with SSR we can cache them a lot longer to save some computation and loading time for our user too. Without JS, at minimum they can still submit the form and get the job done rather than raising a bunch of support ticket to us (I am still not 100% sure if it is achievable with how InertiaJS form work).

docs/cookbook/deploy-with-kamal.md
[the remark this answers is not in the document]
brodienguyen: Thanks @ElMassimo, @skryukov for our detailed discussion and suggestion. I would agree with all those valid points and prepare another PR to allow to set Inertia SSR URL via the ENV variable:
* https://github.com/inertiajs/inertia-rails/pull/172

brodienguyen:
This PR is ready to review. I am still marking it as draft because it depends on another change at:
* https://github.com/inertiajs/inertia-rails/pull/172

docs/cookbook/deploy-with-kamal.md
```diff
@@ -207,19 +207,10 @@ export default defineConfig({
...
-```ruby
-InertiaRails.configure do |config|
-  config.ssr_enabled = ViteRuby.config.ssr_build_enabled
-  config.ssr_url = "#{ViteRuby.config.protocol}://#{ViteRuby.config.host}:13714" # [!code ++]
-  config.version = ViteRuby.digest
-end
-```
+the correct SSR server URL in the adapter. It can be set via the `INERTIA_SSR_URL` ENV variable.
+
+Given we are using kamal 2, I would suggest to leverage [kamal secret](https://kamal-deploy.org/docs/commands/secrets/)
```
skryukov, line 212: Why so? URL without passwords in it isn't a secret at all 😄

brodienguyen: It's true. Ignore my overthinking bro 😂

PR #168, opened 2024-11-28
Fix to_hash called on errors raises an error

### LINKED ISSUE #166: [BUG] NoMethodError for Plain String inside inertia error props in redirects
## Bug Description

After upgrading to **`inertia-rails` 3.4.0**, an error occurs in our application that wasn't present in version 3.3.0. The error seems related to how `inertia` expects errors props.

### Error Details
When attempting to redirect with a plain string as the `errors` value, the following error is raised: 

```
  1) SessionsController POST /sign_in with invalid credentials returns unauthorized status
     Failure/Error: redirect_to sign_in_path, inertia: { errors: "Credenciais Inválidas" }
     
     NoMethodError:
       undefined method `to_hash' for an instance of String
     # ./app/controllers/sessions_controller.rb:22:in `create'
     # ./spec/requests/sessions_controller_spec.rb:26:in `block (4 levels) in <top (required)>'
```

### Steps to Reproduce
1. Upgrade `inertia-rails` from version **3.3.0** to **3.4.0**.
2. Add a `redirect_to` call in your controller that passes a plain string in the `inertia` data, as shown below:
   ```ruby
   redirect_to sign_in_path, inertia: { errors: "Credenciais Inválidas" }
   ```
3. Run the code or corresponding tests.

### Relevant Code Snippet
Here's the controller action causing the error:
```ruby
def create
  user = User.find_by(email: params[:email])

  if user&.authenticate(params[:password])
    @session = user.sessions.create!
    cookies.signed.permanent[:session_token] = { value: @session.id, httponly: true }
    redirect_to logged_path(user)
  else
    redirect_to sign_in_path, inertia: { errors: "Credenciais Inválidas" }
  end
end
```

### Expected Behavior
In version **3.3.0**, passing a plain string in the `errors` key worked without issues. It’s expected that `inertia` should continue supporting plain strings as valid values in the `inertia` data payload.

### Actual Behavior
In version **3.4.0**, it appears `inertia` now expects the `errors` value to be a hash. Passing a plain string raises a `NoMethodError`.

skryukov:
Fixes #166 and adds a deprecation warning to prepare for requiring hashes in a future major update.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -33,8 +33,17 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: ['3.1', '3.2', '3.3']
-        rails: ['6.1', '7.0', '7.1', '7.2']
+        ruby: ['3.0', '3.1', '3.2', '3.3']
+        rails: ['6.1', '7.0', '7.1', '7.2', '8.0']
+        exclude:
+          - ruby: '3.0'
+            rails: '8.0'
+          - ruby: '3.1'
+            rails: '8.0'
+          - ruby: '3.0'
+            rails: '7.2'
+          - ruby: '3.1'
+            rails: '7.2'
 
     runs-on: ubuntu-latest
     name: Test against Ruby ${{ matrix.ruby }} / Rails ${{ matrix.rails }}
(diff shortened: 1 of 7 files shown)


.github/workflows/push.yml
```diff
@@ -54,11 +63,14 @@ jobs:
         uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby }}
+          # Use the latest version of RubyGems with Ruby 3.0 to avoid:
+          # https://bugs.ruby-lang.org/issues/19371
+          rubygems: ${{ startsWith(matrix.ruby-version, '3.0') && 'latest' || 'default' }}
```
bknoles, line 68: i thought i was going crazy for a second because I was sure I'd already merged something like this in... but that was in #162 that targeted the `v2` branch 😅

skryukov: Yup yup I decided it's good to have it in master as well 😄

PR #172, opened 2024-12-04
Allow to set Inertia SSR URL via the ENV variable

brodienguyen:
This PR introduces the ability to set `ssr_url` dynamically using an ENV variable (`INERTIA_SSR_URL`) with a fallback value. This change is based on recent discussions about improving flexibility in configuring the Inertia SSR URL at:
* https://github.com/inertiajs/inertia-rails/pull/167


BrandonShar (approved):
This makes sense to me! The only mild pushback I'd have is that couldn't we already do this with an initializer? The convenience here is nice though.

PR #194, opened 2025-02-16
renderer returns nil when the format is neither html nor x-inertia based

dsh0416:
The injected renderer returns nil when the format is neither html nor x-inertia based, so that the developer might probably reuse the controller for other formats including json or xml api servers.

### DIFF
--- a/lib/inertia_rails/renderer.rb
+++ b/lib/inertia_rails/renderer.rb
@@ -44,6 +44,7 @@ module InertiaRails
         @render_method.call json: page.to_json, status: @response.status, content_type: Mime[:json]
       else
         return render_ssr if configuration.ssr_enabled rescue nil
+        return nil unless @request.format.html?
         @render_method.call template: 'inertia', layout: layout, locals: view_data.merge(page: page)
       end
     end
--- a/lib/inertia_rails/renderer.rb
+++ b/lib/inertia_rails/renderer.rb
@@ -43,8 +43,8 @@ module InertiaRails
         @response.set_header('X-Inertia', 'true')
         @render_method.call json: page.to_json, status: @response.status, content_type: Mime[:json]
       else
-        return render_ssr if configuration.ssr_enabled rescue nil
         return nil unless @request.format.html?
+        return render_ssr if configuration.ssr_enabled rescue nil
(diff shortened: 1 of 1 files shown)


skryukov:
@dsh0416 thanks for the PR, it looks really interesting, could you add a test to support this change (and also demonstrate the cool pattern of reusing the same controller action for inertia and rest api 😎)

dsh0416:
in our use cases, we would first override the `inertia_view_assigns` method with:

```ruby
def inertia_view_assigns
  return {} unless lookup_context.exists?(action_name, controller_path)

  # it looks weird, but it works
  JSON.parse(render_to_string(formats: [ :json ]))
end
```

so that, the controller would automatically search for `jbuilder` files corresponds to the proper action under `app/views`. Similarly, the data retrieved by the inertia views is almost identical to which a RESTful API provides. So that we no longer needs a separate API controller definition in most cases.

I would somehow provide a test case on how this patch responds to raw JSON requests instead of the original ones.

bknoles:
Interesting! Would it be possible to ~~abuse~~ use format responders? Something like:

```ruby
def index
  @products = Product.all
  
  respond_to do |format|
    # 1. aliases HTML format 
    # 2. middleware sets the inertia format for requests with Inertia headers
    format.inertia {
      render inertia: 'Products/Index', props: { products: @products }
    }
    format.json # jbuilder (or whatever) takes over
  end
end
```

dsh0416:
`format.inertia` sounds like an interesting idea, I think it might worth some further edits to the middleware.

bknoles:
Actually, what you want is already possible using respond_to, since both HTML and X-Inertia requests are sent in html format.

```ruby
def index
  @products = Product.all
  
  respond_to do |format|
    format.html { render inertia: 'Products/Index', props: { products: @products } }
    format.json # renders the jbuilder template
  end
end
```

That should work. We've actually got an [old spec](https://github.com/inertiajs/inertia-rails/blob/master/spec/inertia/request_spec.rb#L86) showing this behavior. [Here's the controller](https://github.com/inertiajs/inertia-rails/blob/master/spec/dummy/app/controllers/inertia_test_controller.rb#L80) for it.

bknoles:
Closing because I believe the use case is already covered!

PR #196, opened 2025-02-18
Support configuration via env

skryukov:
This PR generalizes and improves upon the implementation from #172.

I would love to see a more general approach to control configuration values (e.g., `ssr_enabled`)

Having the first point in mind, I needed to choose a strategy to prioritize gem defaults, global user configuration, env options, and finally controller-level configuration.

This is what I came up with, trying to follow the principle of least surprise:
InertiaRails.defaults → ENV → InertiaRails.configure → controller-level config

Example:

```ruby
# initializer
InertiaRails.configure do |c|
  c.ssr_enabled = true
  c.ssr_url = 'localhost:1234'
end

# env
ENV['INERTIA_SSR_ENABLED'] = 'false'
ENV['INERTIA_SSR_URL'] = 'vite:1234'

# controller-level
class PublicController < ApplicationController
end

class DashboardController < ApplicationController
  inertia_config(ssr_enabled: false)
end
```

With this configuration, SSR will be  enabled for `PublicController` but disabled for `DashboardController`.


brodienguyen:
I am sold with this approach. Love your work as always bro @skryukov 
Happy to close my PR to focus on pushing this approach forward.

BrandonShar:
> Usually ENV values take precedence over values set in the initializer

Is this the most common pattern? I would've guessed that manually set options in an initializer should override the environment because they're explicitly set. Everything else makes perfect sense to me!

skryukov:
I think you're right @BrandonShar, that will be more [Rails-y](https://guides.rubyonrails.org/configuring.html#connection-preference), will update the PR.

skryukov:
Yup, it's ready 👍

PR #199, opened 2025-03-01
Allow omitting the component name with props on render

skryukov:
This PR improves the automatic component name functionality, with these changes we can now render a page like this:

```ruby
  def index
    render inertia: { users: User.all.as_json(...) } # => renders pages/users/index.tsx with {users} props
  end
end
```


skryukov:
Linter fails, but I fixed it in other PRs, so will just wait 😂

BrandonShar:
Oh that's nice. Much cleaner than the useless `inertia: true` response. So this is basically for users who want the automatic routes, but still want to be explicit about props?

skryukov:
Yup, used that a lot with `inertia_config default_render: true`

lib/inertia_rails.rb
```diff
@@ -7,6 +7,11 @@
 require 'patches/mapper'
 
 ActionController::Renderers.add :inertia do |component, options|
+  if component.is_a?(Hash) && options[:props].nil?
+    options[:props] = component
+    component = true
+  end
```
bknoles, line 13: Any reason we couldn't/shouldn't put this logic in the renderer file? Its initialize method is already figuring out how to handle the (overloaded) component argument. Might be nice for Future Us to have the argument munging logic in one place.

bknoles:
I think you could do this already to get automatic component paths with explicit props, right?

```ruby
render inertia: true, props: { some: 'stuff' }
```

That said, this does look nicer:

```ruby
render inertia: { some: 'stuff' }
```

I don't love that this `component` argument could now mean 3 things: 1) a component path string, 2) a hash of props, 3) our trusty magic `TrueClass` to tell the controller to go figure out the path on its own, please.

What if we deprecate the `true` option, replacing it (in practice) with:

```ruby
render inertia: {}
```

 in the case where you want the automatic path resolution but don't need to send props.

lib/inertia_rails.rb
bknoles: Should we check for this condition and raise a helpful error?

```ruby
render inertia: { users: User.all }, props: { just_kidding: 'about that users prop!' }
```

As it is I think this would fail as a white screen of death on the frontend?

skryukov: Added a check

skryukov:
@bknoles is there anything holding us from merging this one?

bknoles:
Sorry for the delay! I was on the fence on two things:

1. (warning: this is pedantic) i'd like to rename the `component` argument to `component_or_data` just for future clarity
2. I'd like to deprecate `render inertia: true` in favor of `render inertia: {}`, with an actual deprecation warning (instead of pedantically suggesting `component_or_data_or_magic_true` 😝)

Coming back to it today, neither is important enough to me to delay merging

bknoles (approved):

PR #202, opened 2025-03-22
Fix template detection for scaffolds

skryukov:
This fixes template detection for scaffolds using tailwind 4


BrandonShar (approved):
Nice specs!

PR #207, opened 2025-04-01
Fix 409 for missed pages

### LINKED ISSUE #205: 409 instead of 404 when requesting a non-existent route
I'm getting 409 conflicts instead of 404 errors, when requesting a non-existent route.

I've tracked it down to this line, which usually returns the current asset version, but returns nil when requesting a non-existent route. This results in the 409 conflict: https://github.com/inertiajs/inertia-rails/blob/94bccd9cc0c532d25e31ed92b3301ce87e522cfb/lib/inertia_rails/middleware.rb#L86

skryukov:
This PR fixes #205 plus introduces the `with_inertia_config` test helper and turns on `:random` order for RSpec.


spec/inertia/sharing_spec.rb
```diff
@@ -58,40 +60,36 @@
...
     end
 
     context 'with deep merge configured as the default' do
-      before {
-        InertiaRails.configure { |config| config.deep_merge_shared_data = true }
-      }
-      after {
-        InertiaRails.configure { |config| config.deep_merge_shared_data = false }
-      }
+      with_inertia_config deep_merge_shared_data: true
```
bknoles, line 75: much nicer

bknoles (approved):
looks good!

PR #209, opened 2025-04-07
Add `inertia_rendering?` method and set instance variable in renderer

### LINKED ISSUE #208: Determine whether a view template is being rendered by the `inertia` custom renderer
I noticed that Inertia sets an instance variable `@_inertia_ssr_head` during SSR rendering:

https://github.com/inertiajs/inertia-rails/blob/c8f36c6baa74159415de39459aab1c9fe537d1f6/lib/inertia_rails/helper.rb#L5

We're planning to **gradually** migrate our Rails application to use Inertia, and I need a way to detect in our `application.html.erb` layout whether the current request is being rendered through Inertia. This would help us conditionally include the appropriate JavaScript packs.

Currently, there doesn't seem to be a way to detect if we're in an Inertia rendering context from the application layout (I may be wrong here, but so far I didn't find anything I could use).
 Would it be acceptable if I submit a PR that adds an instance variable (e.g., `@_inertia_rendering`) in the `Renderer` class?

https://github.com/inertiajs/inertia-rails/blob/c8f36c6baa74159415de39459aab1c9fe537d1f6/lib/inertia_rails/renderer.rb#L7-L8

This would allow us to conditionally include assets in our layout like:

```erb
<% if @_inertia_rendering %>
  <%= javascript_include_tag "inertia" %>
<% end %>
```

Andy9822:
We are introducing a new method `inertia_rendering?` in the `InertiaRails` helper to check if the controller is currently rendering an Inertia response.
This allow a view to conditionally render something based on this option i.e. the layout may include the `inertia` pack based on this.

P.S. ideally, this could be subtly mentioned somewhere in the docs but I'm new here so I'll leave that up to you (although Im open to add it if you have a suggestion on where and how to word it if you want)

Closes #208

### DIFF
--- a/lib/inertia_rails/helper.rb
+++ b/lib/inertia_rails/helper.rb
@@ -11,4 +11,8 @@ module InertiaRails::Helper
     )
     inertia_ssr_head
   end
+
+  def inertia_rendering?
+    controller.instance_variable_get("@_inertia_rendering")
+  end
 end
--- a/lib/inertia_rails/renderer.rb
+++ b/lib/inertia_rails/renderer.rb
@@ -37,6 +37,7 @@ module InertiaRails
       @deep_merge = deep_merge.nil? ? configuration.deep_merge_shared_data : deep_merge
       @encrypt_history = encrypt_history.nil? ? configuration.encrypt_history : encrypt_history
       @clear_history = clear_history || controller.session[:inertia_clear_history] || false
+      @controller.instance_variable_set('@_inertia_rendering', true)
     end
 
     def render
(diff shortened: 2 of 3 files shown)


Andy9822:
I'm not completely sure with the testing approach I've implemented for this helper. The current test directly extends the controller with `InertiaRails::Helper`, but this doesn't accurately reflect how the helper is actually included in a Rails application.

In the real application, the helper is included indirectly through `InertiaRails::Controller`, which is then injected into `ActionController::Base` via Rails' `on_load` hook during initialization. This inclusion chain (Controller → Helper) is an important part of how the gem works.

bknoles:
Merged! I'd like to release after we get the deep merge stuff merged in!

Issue #177, opened 2024-12-19
Conflicting type and usage for useForm->errors


mpressen:
in /node_modules/@inertiajs/react/types/useForm.d.ts we can see that :
`errors: Partial<Record<keyof TForm, string>>`

But in the docs, we have this usage :
`redirect_to new_user_url, inertia: { errors: user.errors }` 

which actually create errors with type `Partial<Record<keyof TForm, string[]>>` instead.

Should we do something about it ?

pricci1:
Maybe in Laravel it is a `string`?

Anyways, I added this [patch](https://github.com/ds300/patch-package) for now

```
diff --git a/node_modules/@inertiajs/react/types/useForm.d.ts b/node_modules/@inertiajs/react/types/useForm.d.ts
index de427dc..7cee9a6 100644
--- a/node_modules/@inertiajs/react/types/useForm.d.ts
+++ b/node_modules/@inertiajs/react/types/useForm.d.ts
@@ -7,7 +7,7 @@ type FormOptions = Omit<VisitOptions, 'data'>;
 export interface InertiaFormProps<TForm extends FormDataType> {
     data: TForm;
     isDirty: boolean;
-    errors: Partial<Record<keyof TForm, string>>;
+    errors: Partial<Record<keyof TForm, string[]>>;
     hasErrors: boolean;
     processing: boolean;
     progress: Progress | null;
```

skryukov:
Another option is to define a global re-export:

```ts
// app/frontend/types/inertia-rails.d.ts

import type { FormDataConvertible } from "@inertiajs/core"
import type { InertiaFormProps as OriginalProps } from "@inertiajs/react"

type FormDataType = Record<string, FormDataConvertible>

declare module "@inertiajs/react" {
  interface InertiaFormProps<TForm extends FormDataType>
    extends Omit<OriginalProps<TForm>, "errors" | "setError"> {
    errors: Partial<Record<keyof TForm, string[]>>
    setError(field: keyof TForm, value: string[]): void
    setError(errors: Record<keyof TForm, string[]>): void
  }

  export { InertiaFormProps }

  export function useForm<TForm extends FormDataType>(
    initialValues?: TForm,
  ): InertiaFormProps<TForm>
  export function useForm<TForm extends FormDataType>(
    rememberKey: string,
    initialValues?: TForm,
  ): InertiaFormProps<TForm>
}
```

[...]

PR #215, opened 2025-04-11
Cookbook: Handling Rails validation error types

skryukov:
I believe this should address https://github.com/inertiajs/inertia-rails/issues/177

We could try to upstream the ability to set the `errors` type to the JS core, enabling something like:

```typescript
const form = useForm<UserType, string[]>(...)
```

However, this approach is verbose, and I would still lean towards using global types 😅.

### DIFF
--- a/docs/.vitepress/config.mts
+++ b/docs/.vitepress/config.mts
@@ -154,6 +154,15 @@ export default defineConfig({
                 { text: 'shadcn/ui', link: '/cookbook/integrating-shadcn-ui' },
               ],
             },
+            {
+              text: 'Troubleshooting',
+              items: [
+                {
+                  text: 'Rails validation error types',
+                  link: '/cookbook/handling-validation-error-types',
+                },
+              ],
+            },
           ],
         },
       ],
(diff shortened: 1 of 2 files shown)


pricci1:
Considering that the code [examples](https://github.com/inertiajs/inertia-rails/blob/4ceb8c357f6bfa7acc5b9c9ced7030cfbfde53b3/docs/guide/validation.md?plain=1#L17) in the documentation and the [generators](https://github.com/inertiajs/inertia-rails/blob/4ceb8c357f6bfa7acc5b9c9ced7030cfbfde53b3/lib/generators/inertia/scaffold_controller/templates/controller.rb.tt#L50) pass the errors as arrays, maybe the global re-export should be part of the setup guide or even the `inertia:install` [generator](https://inertia-rails.dev/guide/server-side-setup#rails-generator).

And also, maybe make the type be a string or a string array, though that would force everyone to check the runtime type

```diff
-    errors: Partial<Record<keyof TForm, string>>;
+    errors: Partial<Record<keyof TForm, string | string[]>>;
```

BrandonShar (approved):
I like this strategy of updating docs before trying a more brittle solution!

PR #217, opened 2025-04-14
Fix install generator for typescript setups

Andy9822:
We are now checking if the install generator is running with the `typescript` flag. If so, we'll attempt to replace `<%= vite_javascript_tag 'application' %>` with `<%= vite_typescript_tag 'application' %>`


lib/generators/inertia/install/install_generator.rb
```diff
@@ -125,6 +125,13 @@ def install_typescript
           return
         end
 
+        if File.exist?(application_js_path) && application_layout.read.include?("<%= vite_javascript_tag 'application' %>")
```
Andy9822, line 128: If we can’t find the `application_js_path`, we skip making any changes.
This likely means the user has customized their setup or didn’t follow the full generator path — so we avoid modifying anything to prevent interfering with their configuration.

Andy9822:
hey @bknoles, in case this PR looks good to you, could you help merging this?

I'm not so familiar with contributing to OSS so Im sorry for directly mentioning you, but Im not sure how to properly handle these cases when it comes down to opening a PR, assign or mention someone etc...

BrandonShar:
Thanks for the PR @Andy9822 ! @skryukov knows the generators better than Brian and I so I'd like to give him a chance to weigh in.

skryukov:
@Andy9822 thanks for the PR!

I wonder what happens if a user installs Inertia Rails into an existing app that already uses `application.js` without TypeScript? Perhaps we could move this logic to the Vite installation step to ensure safety.

Andy9822:
> @Andy9822 thanks for the PR!
> 
> I wonder what happens if a user installs Inertia Rails into an existing app that already uses `application.js` without TypeScript? Perhaps we could move this logic to the Vite installation step to ensure safety.

hmm, not sure Im following what issue you're pointing out here @skryukov  🤔 

The code of this PR runs inside the `install_typescript ` function, so the `application.js` would only be converted to `application.ts` when installing Inertia **with TypeScript**

image

I personally don’t think we need to take an "extreme" or so opinionated measure like deleting the application.js file if it’s generated by Inertia — as @bknoles suggested — since it’s actually a fine and useful part of the install command, which helps scaffold a basic structure for you. It’s similar to what you get when generating a new Rails app with its install command.

skryukov:
> hmm, not sure Im following what issue you're pointing out here @skryukov 🤔

I'm talking about a case when someone installs Inertia Rails on top of an existing application (not a `rails new` one) that uses Ruby Vite—in that case, we don't want to touch their files. So, I believe we should either do nothing or move this logic to the `install_vite` step.

skryukov (approved):

PR #219, opened 2025-04-14
Add patch for validation error types on inertia:install

pricci1:
This is part of https://github.com/inertiajs/inertia-rails/pull/215

As [commented](https://github.com/inertiajs/inertia-rails/pull/215#issuecomment-2797236643) in that PR, this changes add the suggested `inertia-rails.d.ts` type overrides to the `inertia:install` generator.

> Considering that the code [examples](https://github.com/inertiajs/inertia-rails/blob/4ceb8c357f6bfa7acc5b9c9ced7030cfbfde53b3/docs/guide/validation.md?plain=1#L17) in the documentation and the [generators](https://github.com/inertiajs/inertia-rails/blob/4ceb8c357f6bfa7acc5b9c9ced7030cfbfde53b3/lib/generators/inertia/scaffold_controller/templates/controller.rb.tt#L50) pass the errors as arrays, maybe the global re-export should be part of the setup guide or even the inertia:install [generator](https://inertia-rails.dev/guide/server-side-setup#rails-generator).

Also, as suggested by @skryukov, a link to the yet-to-be-merged cookbook is included in the `inertia-rails.d.ts` file.

### DIFF
--- a/lib/generators/inertia/install/frameworks.yml
+++ b/lib/generators/inertia/install/frameworks.yml
@@ -16,6 +16,7 @@ react:
     "tsconfig.app.json": "tsconfig.app.json"
     "tsconfig.node.json": "tsconfig.node.json"
     "vite-env.d.ts": "%{js_destination_path}/vite-env.d.ts"
+    "inertia-rails.d.ts": "%{js_destination_path}/types/inertia-rails.d.ts"
   copy_files_js:
     "InertiaExample.jsx": "%{js_destination_path}/pages/InertiaExample.jsx"
   copy_files:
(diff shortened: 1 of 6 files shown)


skryukov:
Thanks for the PR @pricci1!

I believe @bknoles has some pushback on this approach—maybe we should include the `inertia_errors` helper instead to align with Inertia.js types. I also tend to do that in the ShadCN starter, for example.

Another issue with this approach is the fragile nature of these type patches—that's a major downside to this approach tbh 🤔 

@bknoles wdyt?

pricci1:
I agree that it is quite fragile and another thing to maintain.

Feel free to close!

bknoles:
Yea, I like that we're documenting the "redefine the types" option but I don't want to bake it into the library. That said, I echo @skryukov to say thanks for the PR @pricci1 !

PR #228, opened 2025-05-23
fix: missing command on getting started

iaurg:
Hi

There’s a missing step in the Getting Started tutorial. If you follow the instructions as they are, you’ll get the error:

```sh
Could not find generator 'inertia:install'
```
To fix this, we need to install the inertia_rails gem before running the generator.

Also thanks to @kengreeff for the [video](https://www.youtube.com/watch?v=lTSiqGRvuPo)


kengreeff:
Epic work! @iaurg

PR #236, opened 2025-06-20
generate llms full dynamically

BrandonShar:
It's the LLM's world, we're just living in it now 😄


docs/scripts/extract-docs.js
```diff
@@ -0,0 +1,340 @@
+const fs = require('fs')
```
BrandonShar, line 1: I'm not too proud to admit that an LLM wrote this entire file.

BrandonShar:
I'm going to ignore the linting error since it's unrelated, let's fix it in a separate PR.

PR #239, opened 2025-06-21
Deprecate `use_inertia_instance_props`

skryukov:
While `use_inertia_instance_props` is a clever way to mimic Rails behavior, it also exposes too much data on the frontend side, which is very dangerous. Therefore, I propose deprecating it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,54 +41,6 @@ def index
 end
 ```
 
-#### Rails Component and Instance Props 
-
-Starting in version 3.0, Inertia Rails allows you to provide your component name and props via common rails conventions. 
-
-```ruby
-class EventsController < ApplicationController
-  use_inertia_instance_props
-
-  def index
-    @events = Event.all
-  end
-
-end
-```
-
-is the same as 
-
-
-```ruby
-class EventsController < ApplicationController
-  def index
-    render inertia: 'events/index', props: {
-      events: Event.all
-    }
-  end
-end
-```
-
-#### Instance Props and Default Render Notes 
-
-In order to use instance props, you must call `use_inertia_instance_props` on the controller (or a base controller it inherits from). If any props are provided manually, instance props
-are automatically disabled for that response. Instance props are only included if they are defined after the before filter is set from `use_inertia_instance_props`.
-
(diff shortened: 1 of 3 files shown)


bknoles:
> exposes too much data on the frontend side

You mean becuase it's calling `.to_json` without any way to configure it (potentially exposing some fields you shouldn't send to the client)? I don't really use that feature myself, so I want to make sure I understand Problem Land.

skryukov:
yup, that's the main one, but I can think of cases when you simply memoize something and inertia passes that to FE as well.

BrandonShar:
I don't use `use_inertia_instance_props` personally, but I've always seen it [as a sharp knife](https://signalvnoise.com/svn3/provide-sharp-knives/).

I'm fine with discouraging the use of it, but (at the moment at least), I don't think the bar is high enough to deprecate it. Especially since it's opt-in.

bknoles:
Ok, I'm following now. So, I'd say something like this isn't super risky and is a sharp knives situation:

```ruby
class MyController
  before_action :check_something_expensive
  use_inertia_instance_props
  
  def index
    @other_thing = OtherThing.find(params[:id])

    render inertia: 'OtherThing/Show'
  end
  
  # non-Inertia action
  def other_action
    # Reason for memoizing
    render json: my_expensive_thing
  end

  protected

  def check_something_expensive
    raise 'Uh oh' if my_expensive_thing.broken?
  end

  def my_expensive_thing
    @_my_expensive_thing ||= ExpensiveCalc.new.calculate
  end
end
```

However this one would worry me:

```ruby
class MyController < ApplicationController
  include AwesomeLLMHelperFromGem
  use_inertia_instance_props

  def index
    @thing = Thing.find(params[:id])

    render inertia: 'Things/Show'
  end
end

# Inside gem
module AwesomeLLMHelperFromGem
  included do
    before_action do
      @_openai_token = OpenAI.authenticate(some_stuff)
    end
  end

  ...
end 
```

That one would (probably suprisingly) ship credentials right into the Inertia response. As I look through the code, `@_inertia_skip_props` seems intended to mitigate the risk, but it's going to depend on the order that the before_actions are registered / order of operations, etc.

PR #240, opened 2025-07-07
Server driven meta tags

bknoles:
This is a feature I've wanted to add for a long time now. We built an ad hoc version of it ~3 years ago and have been using it in prod since then.

The biggest benefit to this implementation: per-page link previews in a React/Vue/Svelte app _without JS SSR required_. 

The [page I added to the docs](https://de18d259.inertia-rails.pages.dev/cookbook/server-managed-meta-tags
) is comprehensive, but as a quick preview, by adding a single line to the Rails layout and a copying a small component client-side, you can now do:

```ruby
class SomeController < ApplicationController
  def show
    @the_thing = Thing.find(params[:id])

    render inertia 'SomePage', props: { thing: @the_thing }, meta: [
      { name: "description", content: @the_thing.description }
    ]
end
```

Inertia Rails will render a meta tag in the server rendered HTML that Inertia.js will pickup and "take over" rendering during Inertia visits.

There's also an `inertia_meta` controller _instance_ method, that lets you share tags across actions:

```ruby
class SomeController < ApplicationController
  before_action :set_default_meta_tags

  def index
     ...
  end

  def show
    ...
  end

  private

  def set_default_meta_tags
    inertia_meta.add([
      { title: "A common title" },
      { name: "description", content: "The fallback description" }
   ])
end
```

Inertia Rails intelligently deduplicates tags, so you can easily set and override defaults. Plus it uses the same `head_key` as Inertia.js if you want/need more control.

## Test checklist/Todos

- [x] Works in a real React app
- [x] Works in a real Vue app
- [x] Works in a real Svelte app
- [x] Works with SSR (even though the purpose of this is to avoid SSR, i wanted to be sure it wasn't _incompatible_)
- [ ] QA in our production web app (in process)


docs/cookbook/server-managed-meta-tags.md
```diff
@@ -0,0 +1,351 @@
...
+
+### Client Side
+
+Copy the following code into your application. It should be rendered **once** in your application, such as in a [layout component
+](/guide/pages#creating-layouts).
+
+:::tabs key:frameworks
+== Vue
+
+```vue
```
bknoles, line 36: I tested all of these components in shiny new Inertia Rails apps. That said, I've never shipped production Vue/Svelte code. Extra eyes would be appreciated

docs/cookbook/server-managed-meta-tags.md
```diff
@@ -0,0 +1,351 @@
...
+    ...
+    <%= inertia_meta_tags %> <!-- Add this inside your <head> tag -->
+  </head>
+</html>
+```
+> [!NOTE]
+> If you have a `<title>` tag in your Rails layout, make sure it has the `inertia` attribute on it so Inertia knows it should deduplicate it. The Inertia Rails generator does this for you automatically.
+
+
+### Client Side
```
bknoles, line 28: I thought about releasing this in a separate package or trying to merge this into Inertia.js core, but decided the path of least resistance was just to add instructions to copy paste the (tiny) component required to make this work.

docs/cookbook/server-managed-meta-tags.md
```diff
@@ -0,0 +1,351 @@
...
+    {/each}
+  {/if}
+</svelte:head>
+
+```
+:::
+
+## Rendering Meta Tags
+
+Tags are defined as plain hashes and conform to the following structure:
```
bknoles, line 186: I looked at the Rails meta_tags gem and Remix.run's meta objects for inspiration. I decided to minimize "magic". The data structure just describes an HTML, with a few exceptions:

* `tag_name` will default to `:meta` if it isn't defined
* `inner_content` is used for non-void tags
* `title` has shortcut syntax

docs/cookbook/server-managed-meta-tags.md
```diff
@@ -0,0 +1,351 @@
...
+
+  def set_meta_tags
+    inertia_meta.add([
+      { title: 'Look at this event!' }
+    ])
+  end
+end
+```
+
+#### The `inertia_meta` API
```
bknoles, line 257: This is basically the API I _wish_ we had built for `inertia_share`. If you look at the commit history, I actually built out an entire implementation with `inertia_meta` as a controller class method, before remembering how much I regret doing that for `inertia_share` 😅

docs/cookbook/server-managed-meta-tags.md
```diff
@@ -0,0 +1,351 @@
...
+  $: voidTags = metaTags.filter(tag => tag.innerContent == null);
+  $: contentTags = metaTags.filter(tag => tag.innerContent != null);
+
+  let ready = false;
+
+  onMount(() => {
+    // Clean up server-rendered tags
+    document.head
+      .querySelectorAll('[inertia]')
+      .forEach(el => el.remove());
```
bknoles, line 147: This is actually a lot simpler implementation than the React/Vue adapters have. Since you don't have to worry about built in `<Head />` components defining their own meta tags you can safely clear all the server rendered tags and re-render them from scratch.

docs/guide/title-and-meta.md
```diff
@@ -2,6 +2,9 @@
 
 Since Inertia powered JavaScript apps are rendered within the document `<body>`, they are unable to render markup to the document `<head>`, as it's outside of their scope. To help with this, Inertia ships with a `<Head>` component which can be used to set the page `<title>`, `<meta>` tags, and other `<head>` elements.
 
+> [!NOTE]
```
bknoles, line 5: Since this isn't a part of the Inertia.js core, I think it's better to talk about this in a cookbook instead of inside the main title/meta page.

lib/generators/inertia/install/install_generator.rb
```diff
@@ -103,7 +103,7 @@ def install_inertia
                              before: '<%= vite_client_tag %>'
           end
 
-          gsub_file application_layout.to_s, /<title>/, '<title inertia>' unless svelte?
+          gsub_file application_layout.to_s, /<title>/, '<title inertia>'
```
bknoles, line 106: The Svelte component from the doc above needs the inertia tag to properly dedup so I removed this.

lib/inertia_rails/meta_tag.rb
```diff
@@ -0,0 +1,100 @@
...
+      return { inner_content: tag_data[:title] } if @is_shortened_title_tag
+      tag_data.deep_symbolize_keys
+    end
+
+    def tag_digest
+      signature = @tag_data.sort.map { |k, v| "#{k}=#{v}" }.join("&")
+      Digest::MD5.hexdigest(signature)[0, 8]
+    end
+
+    def generate_meta_head_key
```
bknoles, line 82: I added this after testing it in a real app based on feedback from our team. It seemed unncessary to have to add `head_key: "description" everywhere when you know you want to dedup `name: "description"` 100% of the time.

We dedup on the the values in `:name`, `:property` and `:http_equiv` automatically.

lib/inertia_rails/meta_tag.rb
```diff
@@ -0,0 +1,100 @@
...
+    private
+
+    def generate_head_key
+      generate_meta_head_key || "#{@tag_name}-#{tag_digest}"
+    end
+
+    def handle_script_content(content)
+      case content
+      when String
+        @raw ? content.html_safe : ERB::Util.html_escape(content)
```
bknoles, line 56: There's an (undocumented) escape hatch in case you want to render eval'able `<script>` tags. It's undocumented because I can't think of a good reason why you'd actually want to render a script tag like this.

lib/inertia_rails/meta_tag_builder.rb
```diff
@@ -0,0 +1,59 @@
...
+      new_tag = InertiaRails::MetaTag.new(**new_tag_data)
+
+      @meta_tags.reject! do |existing_tag|
+        duplicate?(existing_tag, new_tag)
+      end
+
+      @meta_tags << new_tag
+    end
+
+    def duplicate?(existing_tag, new_tag)
```
bknoles, line 54: Most of the duplication _logic_ is in the head_key generation. Here we just check that, with a special second case for the `title` tag.

BrandonShar (commented):
This is awesome! About time you upstreamed it 😉 

Docs and implementation look great at a glance, but I'm pretty tired right now so I want to give it a second look when I have more brainpower to give it a proper review.

lib/inertia_rails/meta_tag.rb
bknoles: Yea I dug back into my thinking and am going to make a couple changes. I don't think we need eval'able scripts after reconsidering it (I was originally thinking about analytics tracking scripts, but that doesn't really belong in this feature). And I need to escape the content on Inertia requests... the browser normally ignores script content created by `dangerouslySetInnerHTML`, but Inertia.js's `<Head>` renderer overrides it and make it eval'able again.

lib/inertia_rails/generators/controller_template_base.rb
```diff
@@ -11,7 +11,7 @@ class ControllerTemplateBase < Rails::Generators::NamedBase
                                         default: Helper.guess_the_default_framework
 
       class_option :typescript, type: :boolean, desc: 'Whether to use TypeScript',
-                                default: Helper.guess_typescript
+                                default: Helper.guess_typescript?
```
bknoles, line 14: note to self, this will conflict with #234

bknoles:
> @bknoles Is it backward-compatible when we use the syntax that omitting the component name with props on render please?
> 
> * [Allow omitting the component name with props on render #199](https://github.com/inertiajs/inertia-rails/pull/199)

Should work just fine! lemme see if i can confirm with a spec

docs/cookbook/server-managed-meta-tags.md
```diff
@@ -0,0 +1,360 @@
...
+<html>
+  <head>
+    ...
+    <%= inertia_meta_tags %> <!-- Add this inside your <head> tag -->
+  </head>
+</html>
+```
+
+> [!NOTE]
+> If you have a `<title>` tag in your Rails layout, make sure it has the `inertia` attribute on it so Inertia knows it should deduplicate it. The Inertia Rails install generator does this for you automatically.
```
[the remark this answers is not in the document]
bknoles: it's actually important! if you have a title tag in your rails layout withou the inertia attribute AND you define a title tag via inertia, you'll end up with duplicate title tags. I discovered this while building the Svelte head component (which i why I tweaked the generator below)

BrandonShar (commented):
Looks good! I initially wondered why this we need this given SSR, but your objection that we should have to use SSR resonates to with me, that should be opt in.

Just a few questions on some of the implementation!

lib/inertia_rails/meta_tag.rb
```diff
@@ -0,0 +1,111 @@
...
+
+      inner_content =
+        if @tag_name == :script
+          tag_script_inner_content(data.delete(:inner_content))
+        else
+          data.delete(:inner_content)
+        end
+
+      if UNARY_TAGS.include? @tag_name
+        tag_helper.public_send(@tag_name, **data.transform_keys { |k| k.to_s.tr('_', '-').to_sym })
```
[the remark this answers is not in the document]
bknoles: this isn't transforming the UNARY_TAGS, it's other data attributes passed in. so, something like `data_my_attribute` becoming `data-my-attribute`.

lib/inertia_rails/renderer.rb
```diff
@@ -52,6 +52,7 @@ def render
         rescue StandardError
           nil
         end
+        controller.instance_variable_set('@_inertia_page', page)
```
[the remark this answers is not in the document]
bknoles: it won't! the instance props are used in line 31 (above), before this (and the other controller instance variables) are set inside the renderer. i checked, and our existing instance props spec would have caught this if it did impact it!

lib/inertia_rails/meta_tag.rb
```diff
@@ -0,0 +1,111 @@
...
+    def tag_digest
+      signature = @tag_data.sort.map { |k, v| "#{k}=#{v}" }.join('&')
+      Digest::MD5.hexdigest(signature)[0, 8]
+    end
+
+    def generate_meta_head_key
+      return unless @tag_name == :meta
+      return 'meta-charset' if @tag_data.key?(:charset)
+
+      %i[name property http_equiv].each do |key|
```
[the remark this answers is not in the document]
bknoles: sure! done!

PR #254, opened 2025-08-08
[Feature] Add a configuration option to transform props coming from Inertia

### LINKED ISSUE #247: [Suggestion] Add support for custom prop transformers
## Problem statement

Some people (like myself) greatly prefer using all camelCase in JS and all snake_case in Ruby. Inertia doesn’t support this paradigm by default, causing the developer to use one casing scheme in both enviroments.

## Proposed fix

Maybe Inertia could add a configuration option for a prop transformer - a lambda that runs before Inertia renders the response to the DOM. Something like this:

```ruby
    inertia_config(
      prop_transformer: ->(props:) do
        props.deep_transform_keys! { |key| key.to_s.camelize(:lower) }
      end
    )
```

This is just an example - other considerations may need to be made for the other data processed by `InertiaRails::Renderer`. Note that this doesn’t handle transforming the props coming into Rails and I’m not sure if it should - I’m on the fence.

## Workaround

This is what I’ve been using for a few weeks with good success. It may not cover all cases, but it’s been working for me:

```ruby
# config/initializers/inertia_rails.rb

# [Other config]

module InertiaRails
  class CustomRenderer < InertiaRails::Renderer
    private

    def deep_transform_props(props, parent_path = [])
      super.deep_transform_keys! { |key| key.to_s.camelize(:lower) }
    end
  end
end

# This is needed to transform props keys to camelCase which is a personal
# preference when consuming Inertia props in Vue components.
ActionController::Renderers.add(:inertia) do |component, options|
  InertiaRails::CustomRenderer.new(
    component,
    self,
    request,
    response,
    method(:render),
    props: options[:props],
    view_data: options[:view_data],
    deep_merge: options[:deep_merge],
    encrypt_history: options[:encrypt_history],
    clear_history: options[:clear_history]
  ).render
end
```

Then, to transform props coming _into_ Rails, I use this `before_action` globally via `ApplicationController`:

```ruby
  before_action :underscore_params

  # …

  private

  def underscore_params
    params.deep_transform_keys! { |key| key.to_s.underscore }
  end
```

kieraneglin:
Resolves #247

Let me know what you think!


docs/guide/configuration.md
```diff
@@ -44,6 +44,33 @@ Inertia Rails supports setting any configuration option via environment variable
...
+
+```ruby
+before_action :underscore_params
+
+# ...
+
+def underscore_params
+  params.deep_transform_keys! { |key| key.to_s.underscore }
+end
+```
```
kieraneglin, line 72: This probably needs work. Open to suggestions!

lib/inertia_rails/renderer.rb
```diff
@@ -90,10 +90,13 @@ def merge_props(shared_props, props)
...
-      deep_transform_props(merged_props).tap do |transformed_props|
-        transformed_props[:_inertia_meta] = meta_tags if meta_tags.present?
-      end
+      merge_props(shared_data, props)
+        # This performs the internal work of hydrating/filtering props
+        .then { |props| deep_transform_props(props) }
+        # Then we apply the user-defined prop transformer
+        .then { |props| configuration.prop_transformer(props: props) }
+        # Then we add meta tags after everything since they must not be transformed
+        .tap { |props| props[:_inertia_meta] = meta_tags if meta_tags.present? }
```
kieraneglin, line 99: This order ensures that any current or future internal props (like `_inertia_meta`) are untouched by the user-provided transformer. As long as they're added later in the chain than the call to `prop_transformer`, they're good to go

spec/dummy/app/controllers/inertia_prop_transformer_controller.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+class InertiaPropTransformerController < ApplicationController
```
kieraneglin, line 3: I was kind of winging it with this. Open to feedback!

docs/guide/configuration.md
skryukov: I would drop this additional section about input props tbh, this configuration option is not only about casing, so it looks misplaced to me

bknoles (approved):
This looks good to me! I'll wait to merge it until right before I release since the docs updates will build to the docs site immediately

PR #257, opened 2025-08-13
Fix `ActionView::MissingTemplate` for static controller

### LINKED ISSUE #248: ActionView::MissingTemplate with InertiaRails::StaticController
When using the `InertiaRails::StaticController` we noticed that we receive `ActionView::MissingTemplate` exceptions. We fixed the same issue in our controllers by wrapping the `render` call into a `respond_to` block like this:

```ruby
respond_to do |format|
   format.html do
     render inertia: "..."
  end
end
```

So I think this could be fixed by doing the same in `InertiaRails::StaticController`. What do you think? Are there other formats that inertia could theoratically respond to?

skryukov:
There are two ways to address the `ActionView::MissingTemplate` issue: using a `respond_to` block or forcing the `:html` format.

After clarification from @richardboehme, we decided to go with `respond_to` 😂 

Fixes #248

### DIFF
--- a/app/controllers/inertia_rails/static_controller.rb
+++ b/app/controllers/inertia_rails/static_controller.rb
@@ -1,7 +1,9 @@
 module InertiaRails
   class StaticController < InertiaRails.configuration.parent_controller.constantize
     def static
-      render inertia: params[:component]
+      respond_to do |format|
+        format.html { render inertia: params[:component] }
+      end
     end
   end
 end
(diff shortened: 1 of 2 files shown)


bknoles:
I played with this a bit today. I'm in favor of wrapping the renderer in a `repond_to` block.

IMO the issue is that we're raising a _confusing_ exception.

With a `respond_to` block constraining us to html formats, Rails raises an UnknownFormat error, which I think is a helpful exception. Since there are still ways the app can decide to support `/some_inertia_path.some_other_format` if it wants to, coercing it in the renderer is overly prescriptive for me. What do you think?

bknoles (approved):

PR #260, opened 2025-08-21
Update errors params handling

skryukov:
This PR addresses multiple things:
- makes sure `errors` params are always returned (important for cases when partial request returns errors)
- introduces a new config that makes errors hash always being present in the response (with an option to turn it off) See the protocol for more info: https://inertia-rails.dev/guide/the-protocol#the-page-object
- updates generator's initializer template


bknoles:
This looks good to me! I like the existing-app-friendly deprecation path.

Could you add a quick spec for the always return errors case? It'd be nice to have that documented in the specs

PR #263, opened 2025-09-01
Complete rubocop todo

skryukov:
TODO is something we should actually do. This PR fixes all RuboCop offenses.


lib/patches/debug_exceptions.rb
```diff
@@ -1,17 +1,34 @@
...
 
-if defined?(ActionDispatch::DebugExceptions)
-  if ActionPack.version.to_s >= '5.1'
-    require 'patches/debug_exceptions/patch-5-1'
-  elsif ActionPack.version.to_s >= '5.0'
-    require 'patches/debug_exceptions/patch-5-0'
-  else
-    # This gem supports Rails 5 or later
+module InertiaRails
+  module InertiaDebugExceptions
```
skryukov, line 14: Dropped the patch for Rails < 5.1 since Inertia Rails requires Rails 6+

Issue #266, opened 2025-09-15
Inertia configuration mismatch when custom `prop_transformer`


pedrofs:
I'm doing `prop_transformer` to maintain naming conventions between React and Rails:
```ruby
  inertia_config prop_transformer: ->(props:) do
                   # We need to parse to JSON and parse back to force ActiveRecord::Base instances to be serialized
                   # otherwise `deep_transform_keys` won't work on their attributes
                   serialized_props = JSON.parse(props.to_json, symbolize_names: true)

                   serialized_props.deep_transform_keys { |key| key.to_s.camelize(:lower) }
                 end
```

I'm seeing an issue now with merge and deep merge functionality. The deep merge response result comes with snake_case while the prop is camelCase:
<img width="645" height="120" alt="Image" src="https://github.com/user-attachments/assets/e8bc7116-016f-4a78-b41e-ec6cc7a11746" />

I was wondering if we could expose a similar feature as `prop_transformer` to expose a transformer for the merges. Or, is there a better way to solve this inconsistency?

Thanks!

bknoles:
Well the entire purpose of `prop_transformer` from #254 is to modify keys. I agree that having the props concept split up into `props`, `mergeProps`, `deepMergeProps`, as Inertia.js now has, makes that feature (and anything touching prop rendering) more awkward.

But short of removing `prop_transformer` altogether, I'd say that just doing the shotgun surgery is an ok path forward.

(fwiw, my team doesn't use this feature... we serialize with Blueprinter and camel case keys in a transformer)

pedrofs:
Hey guys, thanks for the reply!

I agree we should keep the prop transformer functionality, but we also need to support it across the board.

I'm happy to apply the same transformation to all the places it needs to be if you guys agree to merge the PR afterwards.

@bknoles , @skryukov thoughs?

PR #267, opened 2025-09-15
Implement merge_prop_transformer configuration

pedrofs:
I'm using the `prop_transformer` to maintain a clear naming convention between React and Rails: React is camelCase and Rails is snake_case.

Unfortunately, I faced an inconsistency when trying to work with the InertiaRails.merge functionality. See #266 .

This PR introduces the `merge_prop_transformer` configuration, similar to the `prop_transformer`. It'll receive the array of merge keys, and you can transform them as you wish.

### DIFF
--- a/lib/inertia_rails/renderer.rb
+++ b/lib/inertia_rails/renderer.rb
@@ -135,7 +135,9 @@ module InertiaRails
         default_page[:deepMergeProps] = configuration.merge_prop_transformer(merge_props: deep_merge_props.map(&:first))
       end
 
-      default_page[:matchPropsOn] = match_props_on if match_props_on.present?
+      if match_props_on.present?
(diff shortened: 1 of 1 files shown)


bknoles:
We've talked about this and decided that the JS side is the better place to do key transformation. That way we avoid having to worry about transforming the keys that come _back_ to the server from the React app @skryukov built an NPM packages that handles converting to camelCase for Inertia apps that does this all very cleanly:

https://github.com/skryukov/inertia-caseshift

We'd recommend using that instead of transforming props in Rails.

We're going to deprecate the Inertia Rails prop transformation feature (instead of improving it) and eventually remove it.

That said, thanks for submitting the PR and sorry for this sitting around for so long!

PR #277, opened 2025-11-09
Fix for shared scroll and deferred props

bknoles:
Found an issue with our newly released infinite scroll support! Scroll and deferred prop metadata doesn't render to the client if they are shared data. The reason is that they were being computed by iterating over `@props`, which doesn't include shared data.

I moved the merging of controller props and shared props into the initializer. We don't need them separately anywhere in the renderer and seems like Future Us might assume we can iterate over `@props` someday.

The renderer has become a bit of a maze as it has grown, and one thing that made it tricky to look through is that the attr_readers are used to reference instance variables... but only sometimes. And some methods use those method names as arguments. I think it's cleaner to just consistently reference the instance variables themselves.

### DIFF
--- a/lib/inertia_rails/renderer.rb
+++ b/lib/inertia_rails/renderer.rb
@@ -6,6 +6,16 @@ require_relative 'inertia_rails'
 
 module InertiaRails
   class Renderer
+    %i[component configuration controller props view_data encrypt_history
+       clear_history].each do |method_name|
+      define_method(method_name) do
+        InertiaRails.deprecator.warn(
+          "[DEPRECATION] Accessing `InertiaRails::Renderer##{method_name}` is deprecated and will be removed in v4.0"
+        )
+        instance_variable_get("@#{method_name}")
+      end
+    end
+
     def initialize(component, controller, request, response, render_method, **options)
       if component.is_a?(Hash) && options.key?(:props)
         raise ArgumentError,
--- a/spec/inertia/renderer_spec.rb
+++ b/spec/inertia/renderer_spec.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+RSpec.describe InertiaRails::Renderer do
+  let(:deprecator) do
+    double(warn: nil).tap do |deprecator|
+      allow(InertiaRails).to receive(:deprecator).and_return(deprecator)
+    end
+  end
+
+  %i[component configuration controller props view_data encrypt_history clear_history].each do |method_name|
+    it "has a deprecated #{method_name} accessor" do
+      configuration = double('configuration',
+                             encrypt_history: true,
+                             deep_merge_shared_data: false,
+                             clear_history: false)
+
+      controller = double('controller',
+                          inertia_configuration: configuration,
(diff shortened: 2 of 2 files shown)


lib/inertia_rails/renderer.rb
```diff
@@ -6,16 +6,6 @@
...
   class Renderer
-    attr_reader(
-      :component,
-      :configuration,
-      :controller,
-      :props,
-      :view_data,
-      :encrypt_history,
-      :clear_history
-    )
```
bknoles: These are public but they aren't referenced anywhere else in the codebase and I can't imagine that someone is calling them in the wild somewhere. Please someone sanity check me on that!

lib/inertia_rails/renderer.rb
```diff
@@ -24,15 +14,19 @@ def initialize(component, controller, request, response, render_method, **option
...
-      @props = options.fetch(:props, component.is_a?(Hash) ? component : controller.__send__(:inertia_view_assigns))
       @view_data = options.fetch(:view_data, {})
-      @deep_merge = options.fetch(:deep_merge, configuration.deep_merge_shared_data)
-      @encrypt_history = options.fetch(:encrypt_history, configuration.encrypt_history)
+      @encrypt_history = options.fetch(:encrypt_history, @configuration.encrypt_history)
       @clear_history = options.fetch(:clear_history, controller.session[:inertia_clear_history] || false)
+
+      deep_merge = options.fetch(:deep_merge, @configuration.deep_merge_shared_data)
+      passed_props = options.fetch(:props, component.is_a?(Hash) ? component : @controller.__send__(:inertia_view_assigns))
+      @props = merge_props(shared_data, passed_props, deep_merge)
```
bknoles: This line is the fix... merging props immediately so that downstream methods can reference `@props` and have the full set of props. This seems like the least surprising behavior.

lib/inertia_rails/renderer.rb
```diff
@@ -91,16 +85,15 @@ def merge_props(shared_props, props)
 
     def computed_props
       # rubocop:disable Style/MultilineBlockChain
-      merge_props(shared_data, props)
-        .then do |merged_props| # Always keep errors in the props
+      @props
+        .tap do |merged_props| # Always keep errors in the props
```
bknoles: other part of the fix (just making sure we aren't unnecessarily calling merge_props twice)

lib/inertia_rails/renderer.rb
BrandonShar: I hope you're right, but this is a breaking change right? Could they be used in tests?

BrandonShar: The rspec helpers mock it: https://github.com/inertiajs/inertia-rails/blob/99774c86480f4b4f696d2b41b486565b20be5194/lib/inertia_rails/rspec.rb#L85 but it looks like you're right that they don't care about these particular vars.

bknoles: that's mocking the _arguments_ to `.new` rather than the accessor methods

PR #290, opened 2025-11-22
Add request lifecycle diagram

skryukov:
- Added a new "Request lifecycle diagram" section with a visual diagram
- Updated the page object documentation to include more detailed descriptions
- Revised and expanded example page objects to consistently include the `errors` property

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -4,17 +4,36 @@
   "requires": true,
   "packages": {
     "": {
+      "dependencies": {
+        "vue-tsc": "^3.1.4"
+      },
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


BrandonShar:
What are you thoughts on changing "Returns HTML skeleton" to something referencing that it returns the inertia ERB along with props/view vars? I think that makes it a bit more clear what's happening in a literal sense (which seems to be what we're targeting with the diagram)

bknoles:
Yea, I think this is a great addition as is! Thanks @mattdb for generating the diagram a couple weeks ago :)

bknoles (approved):

PR #291, opened 2025-11-22
Use new tagline & colors

skryukov:
They are awesome


BilalBudhani:
this is awesome! Do you guys think we should match the fonts too?

PR #298, opened 2025-11-27
refactor: improve TypeScript setup for non-interactive installs

alec-c4:
Move TypeScript configuration files (tsconfig.json, tsconfig.node.json, tsconfig.app.json) and type definitions (vite-env.d.ts, globals.d.ts, index.ts) from frameworks.yml to install_generator.rb to ensure they are always copied when --typescript flag is used, regardless of interactive mode.

### DIFF
--- a/lib/generators/inertia/install/install_generator.rb
+++ b/lib/generators/inertia/install/install_generator.rb
@@ -118,7 +118,7 @@ module Inertia
           ERB
           insert_into_file application_layout.to_s, headers, after: "<%= vite_client_tag %>\n"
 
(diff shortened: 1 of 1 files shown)


skryukov:
Hey @alec-c4, could you show repro steps for a bug this PR is addressing?

alec-c4:
@skryukov, sure! when you run `rails generate inertia:install --framework=svelte --typescript --no-interactive --tailwind --vite --no-example-page --force` it doesn't copy tsconfigs and types

skryukov:
oh, I see, good catch, I want to rewrite the whole thing tbh to make it:
1) get the hole config from flags
2) copy all files at once
3) install all deps at once

PR #354, opened 2026-04-06
Skip excluded hash props on partial reloads

erickreutz:
## Summary
While debugging a regression after upgrading to Inertia v3, I found that partial reloads can still traverse excluded hash containers.

When that happens, nested props can leak into the response even though the parent hash was not requested, leaving a partial version of the parent payload instead of excluding it entirely.

This adds the same partial-request guard that arrays already have before recursing into hashes.

The spec covers the smaller repro: an excluded parent hash with a nested `AlwaysProp` should stay excluded instead of leaking back into the response.


skryukov (approved):

skryukov:
Thanks @erickreutz!

PR #356, opened 2026-04-10
Add prop-level caching and caching documentation

skryukov:
This PR adds `InertiaRails.cache` for server-side prop caching, and docs on caching.

### DIFF
--- a/docs/guide/deferred-props.md
+++ b/docs/guide/deferred-props.md
@@ -252,3 +252,21 @@ end
 ```
 
 For more information on once props, see the [once props](/guide/once-props) documentation.
+
+## Combining with Caching
+
+@available_since rails=master
+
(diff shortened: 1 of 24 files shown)


docs/guide/deferred-props.md
```diff
@@ -252,3 +252,21 @@ end
...
+
+@available_since rails=master
+
+You may pass the `cache` option to a deferred prop to cache the resolved value on the server side. On cache hits, the block is not evaluated.
+
+```ruby
+class DashboardController < ApplicationController
+  def index
+    render inertia: {
+      feed: InertiaRails.defer(cache: 'user_feed', group: 'feed') { current_user.feed },
```
bknoles: The example would never expire right? That's totally possible and the user's choice with the Rails cache, but maybe the examples should include expire_in. wdyt?

bknoles (approved):

PR #364, opened 2026-04-13
Skip session cleanup in middleware when session was never loaded

khamusa:
Hello 👋  Thanks for the amazing work on this gem 🙌

## Problem

`InertiaRails::Middleware` unconditionally calls `session.delete` on three Inertia session keys after every response:

```ruby
unless keep_inertia_session_options?(status)
  request.session.delete(:inertia_errors)
  request.session.delete(:inertia_clear_history)
  request.session.delete(:inertia_preserve_fragment)
end
```

`session.delete` calls `load_for_write!` internally, which loads (and in the case of a fresh request, creates) the session regardless of whether the request ever touched it. In apps where Inertia coexists with sessionless controllers (e.g. token-authenticated API endpoints, webhook receivers), this causes the session to be loaded and committed on **every** request, producing unnecessary database writes:

```
SAVEPOINT active_record_1
INSERT INTO "sessions" (...)
RELEASE SAVEPOINT active_record_1
SELECT "sessions".* WHERE "session_id" = $1
```

## Root cause

The three Inertia keys can only exist in the session if they were written there during a prior `redirect_to`. If the session was never loaded during the current request, those keys provably cannot be present — the cleanup is a no-op that still forces session I/O.

## Fix

Merge the `session.loaded?` guard into the outer `unless` condition:

```ruby
unless keep_inertia_session_options?(status) || !request.session.loaded?
  request.session.delete(:inertia_errors)
  request.session.delete(:inertia_clear_history)
  request.session.delete(:inertia_preserve_fragment)
end
```

- If `session.loaded?` is `false`, the Inertia keys cannot have been set — skip entirely.
- If `session.loaded?` is `true`, existing behaviour is preserved exactly.

## Tests

Two new regression tests are added to `middleware_spec.rb`:

1. **Does not load the session** for a request that never accesses it — asserts `request.session.loaded?` remains `false` after the response (would fail against the old code).
2. **Still cleans up inertia session keys** when the session was legitimately loaded during the request — asserts `session[:inertia_errors]` is `nil` after following a redirect that stored errors.

## Documentation

A new **Sessionless Controllers in Hybrid Applications** section is added to the CSRF protection guide.

While investigating this issue we found that a related problem arises when sessionless controllers use `skip_forgery_protection`. `skip_forgery_protection` only removes the `verify_authenticity_token` before-action — it does not disable the CSRF infrastructure. `protect_against_forgery?` still returns `true`, so InertiaRails' after-action in the controller calls `form_authenticity_token` and loads the session on every request.

The correct per-class configuration for sessionless controllers is `self.allow_forgery_protection = false`. This makes `protect_against_forgery?` return `false` unconditionally, so InertiaRails' after-action is skipped entirely — no XSRF cookie, no `form_authenticity_token` call, no session I/O. The setting is scoped to the controller class and does not affect the rest of the application.

> **Note:** An earlier version of this PR recommended `protect_from_forgery with: :null_session` instead. That was a mistake — `null_session` only prevents session I/O on non-GET requests (it works by substituting a `NullSessionHash` on CSRF verification failure, which only fires for non-safe methods). `allow_forgery_protection = false` is the correct tool: it disables the CSRF infrastructure at the class level regardless of request method.

The middleware fix in this PR addresses the remaining session I/O from the cleanup loop for any controller that never loaded the session, regardless of CSRF configuration.

### DIFF
--- a/docs/guide/csrf-protection.md
+++ b/docs/guide/csrf-protection.md
@@ -69,33 +69,29 @@ createInertiaApp({
 })
 ```
 
-## Stateless Controllers in Hybrid Applications
+## Sessionless Controllers in Hybrid Applications
 
 When Inertia coexists with sessionless controllers in the same Rails application — such as token-authenticated API endpoints, webhook receivers, or any controller that does not rely on the session — it's important to configure CSRF protection correctly on those controllers.
 
 A common pattern is to reach for `skip_forgery_protection`:
 
 ```ruby
-class StatelessController < ApplicationController
+class SessionlessController < ApplicationController
   skip_forgery_protection
 end
 ```
 
 However, `skip_forgery_protection` only removes the `verify_authenticity_token` before-action — it does not disable the CSRF infrastructure. Rails' `protect_against_forgery?` still returns `true`, so InertiaRails' after-action fires and calls `form_authenticity_token`, which reads and writes `session[:_csrf_token]`. This causes a session record to be loaded (and created, if one doesn't exist) for every request, even though the controller has explicitly opted out of CSRF.
 
-The correct approach is to use the `:null_session` strategy instead:
+The correct approach is to set `allow_forgery_protection` to `false` on the controller class:
 
 ```ruby
-class StatelessController < ApplicationController
-  protect_from_forgery with: :null_session
+class SessionlessController < ApplicationController
+  self.allow_forgery_protection = false
 end
 ```
 
-With `:null_session`, Rails runs `verify_authenticity_token` but handles the expected absence of a CSRF token by substituting a `NullSessionHash` for the real session. This causes `protect_against_forgery?` to return `false` for the remainder of that request — InertiaRails' after-action is correctly skipped, and no session I/O occurs.
-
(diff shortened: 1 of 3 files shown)


skryukov (approved):
Awesome findings, thank you @khamusa!

khamusa:
@skryukov I'm getting failures because of the documentation changes, can you advise?

skryukov:
Everything is fine, that's an issue with our highly vibed attempt to make documentation changes deployed from forks 😄

khamusa:
Been there 😆 ! Let me know if you need anything from me, and thanks for the super quick turnaround on this 🙌

khamusa:
@skryukov I took this back to draft as some of my assumptions in the updated documentation are wrong. I'm reviewing the whole thing.

khamusa:
@skryukov                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                            
I initially recommended `protect_from_forgery with: :null_session` in the documentation, which was wrong... 😓 that strategy only prevents session I/O on non-GET requests since it works by substituting a NullSessionHash on verification failure, and verification never fires for safe methods. Caught it before merging and corrected it. 🙌             
                                         
The updated recommendation is to set `self.allow_forgery_protection = false` instead.

PR #368, opened 2026-04-14
Add configurable XSRF cookie refresh policy

darkamenosa:
## Summary

Adds a configurable XSRF cookie refresh policy to the Rails adapter:

```ruby
  config.xsrf_cookie_refresh = :lazy
end
```

Supported values:

- `:always` keeps the current behavior and refreshes `XSRF-TOKEN` on every protected request.
- `:lazy` skips the rewrite on `GET` / `HEAD` when the cookie already exists. When the session or CSRF token is already loaded, the existing cookie is validated and refreshed if stale. Non-safe requests continue to refresh on every protected request.

## Why

The adapter refreshes `XSRF-TOKEN` in an `after_action` on every protected request. Because Rails masks CSRF tokens with a random one-time pad on each call (BREACH mitigation), `form_authenticity_token` returns a different value every time — so Rails' built-in same-value `Set-Cookie` suppression never applies, and every response carries a `Set-Cookie` header, including `304 Not Modified` responses.

That defeats HTTP caching for apps that rely on it: most CDNs and shared caches (Cloudflare, Varnish, rack-cache) refuse to cache responses carrying `Set-Cookie`, and browser conditional caching (`fresh_when` / `stale?` → `ETag` / `304`) can fail to settle, observed most notably in Safari/WebKit.

This PR adds an opt-in way to remove that churn on steady-state safe requests without changing the default behavior.

## Design

- Enum policy instead of a boolean — reads clearly and leaves room for a future `:never` value (relevant once Rails 8.2's `Sec-Fetch-Site`-based `:header_only` forgery protection makes the token cookie unnecessary).
- Because the masked cookie value can't be compared by equality, the only correct "did the token change?" test is unmasking via `valid_authenticity_token?`. The `:lazy` path validates opportunistically: whenever the session or CSRF token is already loaded, a stale cookie is detected and refreshed — including on `304` responses, since `fresh_when` loads the session. The adapter never loads a lazy session *just* to validate, because loading a cookie-backed session emits its own session `Set-Cookie` and would reintroduce the same churn.
- Invalid values fail at configuration build time (`InertiaRails.configure`, `inertia_config`, and `INERTIA_XSRF_COOKIE_REFRESH` all raise eagerly).
- Configurable globally, per controller via `inertia_config`, or via `INERTIA_XSRF_COOKIE_REFRESH`.

## Trade-off

On `GET` / `HEAD` requests where nothing loads the session, an existing cookie is trusted without validation. A stale cookie (for example after a `secret_key_base` rotation) is then not detected until a protected request fails with `InvalidAuthenticityToken` — a loud failure, never a silent bypass (the cookie is not a server-side validation input), and it self-heals on the first session-loading request, including every rendered Inertia response. The cookbook page documents this caveat; a regression spec pins the behavior.

## Docs

Updates the configuration guide and CSRF guide, and adds a cookbook page for the HTTP caching use case, including the caveats above and the note that cookie-backed sessions are a separate source of `Set-Cookie` churn this option does not address.

## Validation

- `bundle exec rspec spec/inertia` — full suite green, including new regression specs for: 304 responses under `:always` vs `:lazy`, stale-cookie refresh on 304s, the session-less trust path, and eager config validation.
- CI matrix covers Rails 6.1–8.1; the `CSRF_TOKEN` env-key cleanup is a deliberate no-op below Rails 7.1, where the key does not exist.

## Checklist

- [x] Tests added or updated


lib/inertia_rails/configuration.rb
```diff
@@ -3,6 +3,7 @@
 module InertiaRails
   class Configuration
     DEFAULT_SSR_URL = 'http://localhost:13714'
+    XSRF_COOKIE_REFRESH_OPTIONS = %i[always when_needed].freeze
```
skryukov: I'd replace `when_needed` with `lazy`, a bit less vague naming. Or maybe go with a boolean option i.e. `xsrf_cookie_lazy_refresh`, that will simplify config, wdyt?

lib/inertia_rails/controller.rb
```diff
@@ -124,6 +127,12 @@ def inertia_configuration
       self.class._inertia_configuration.bind_controller(self)
     end
 
+    def skip_xsrf_cookie_refresh?
+      inertia_configuration.xsrf_cookie_refresh == :when_needed &&
+        (request.get? || request.head?) &&
+        request.cookies['XSRF-TOKEN'].present?
```
skryukov: I wonder if calling smth like `valid_authenticity_token?(session, request.cookies['XSRF-TOKEN'])` here might help us to eliminate the outdated cookie risk.

docs/cookbook/http-caching-and-xsrf-cookie-refresh.md
```diff
@@ -0,0 +1,79 @@
+# HTTP Caching and XSRF Cookie Refresh
+
+If your Rails app uses HTTP conditional caching (`ETag` / `Last-Modified` -> `304 Not Modified`) on Inertia pages, you may notice that browser revalidation does not work as expected even though your controller code looks correct.
```
skryukov: iirc, it's not a general issue, more like a Safari quirk. Chrome works just fine with current defaults. I also wonder if we should advise moving session storage off from cookies into Redis/DB since that will also play same effect (haven't checked that though personally, but I believe you have a note on that behavior downwards).

skryukov (commented):
Hey @darkamenosa thanks for the PR! I left some nit-picky comments 😄

darkamenosa:
@skryukov give me a little time. Just stuck with life stuff. I will check these as soon as possible.

docs/cookbook/http-caching-and-xsrf-cookie-refresh.md
darkamenosa: Good point. I updated this as a caveat rather than a guaranteed recommendation.

I checked the Rails path here: with the default CSRF session store, `valid_authenticity_token?` can reach `real_csrf_token`, which fetches from `request.session[:_csrf_token]`. That session read can load an otherwise untouched lazy session, and Rack commits loaded sessions by emitting a session cookie. So if we validated the existing `XSRF-TOKEN` unconditionally on every safe request, we could recreate the same `Set-Cookie` churn via the Rails session cookie.

That is why the lazy behavior now validates opportunistically: if the request has already loaded the session or CSRF token, we validate the existing XSRF cookie and refresh it if stale. But we avoid loading the session only for that validation on otherwise steady-state `GET` / `HEAD` requests.

I also added the Redis/DB session storage point in the docs under the caveats. Moving session storage off cookie-backed sessions can reduce that separate source of cookie churn, but I avoided saying it always fixes the issue because Rails may still emit a session-id cookie when the request loads or mutates session state.

darkamenosa:
@skryukov I'm not sure, if this should be added to inertia. Since the goal is trying to make caching work. 

304 response made huge improvement on speed. But inertia current caching is not so good like normal .html.erb (or I don't know how to do things properly), so I have to do a lot of ad hoc things to make it works. (cache using rails cache, 304 response, cloudflare cache).

Here is a site I applied above stuffs to improve the speed. https://contextqmd.com/ - The DB is about 14GB, with 5.5m rows, on $5, Hetzner server. 

If there is better way, pls suggest for me.

skryukov:
@darkamenosa I think this PR makes a lot of sense, will review it once again soon and merge. Meanwhile have you seen InertiaRails.cache from the last release? Would love to hear your thoughts on it https://github.com/inertiajs/inertia-rails/pull/356

darkamenosa:
@skryukov ok, let me check it out

darkamenosa:
@skryukov I've just read the docs. It seems complete. Also covered the SSR cache, this also a thing I wanted to create a pull request. Let me update the version and try it out. 

Read the docs, I saw the new puma ssr plugin. This is great. Question, will you support falcon for something similar? Nowaday, calling LLM request (long response time) puma is not enough. That's the reason I have to switch to falcon.

skryukov:
re puma, I guess it's possible, according to https://github.com/socketry/falcon/discussions/279

darkamenosa:
@skryukov That's great. Also there is another problem, I'm not sure should do an upstream patch. Let me explain. 

There may be a small HTTP caching gap around `Vary: X-Inertia` and Rails conditional GET.

`inertia_rails` currently sets `Vary: X-Inertia` inside `InertiaRails::Renderer#render`, which works for normal `200` responses. However, with Rails `stale?` / `fresh_when`, Rails can short-circuit and return `304 Not Modified` before `render inertia:` runs.

Flow:
```
Controller action
  |
  | stale? / fresh_when
  |
  | if fresh: Rails returns 304 here
  |           render inertia never runs
  |
  v
render inertia
  |
  | InertiaRails::Renderer adds Vary: X-Inertia
```

So for 304 responses, the renderer-level Vary logic can be too late.

This matters because the same URL can return either HTML or an Inertia JSON page object depending on the request header:
```
GET /posts/1
# HTML

GET /posts/1
X-Inertia: true
# JSON
```
etag { request.inertia? } solves only the validator split. It makes HTML and JSON produce different ETags.

But Vary: X-Inertia solves the cache variant split. It tells browsers/CDNs/proxies that X-Inertia participates in choosing the representation. Without it on conditional responses, a cache can potentially associate/revalidate the wrong HTML/JSON variant for the same URL.

Locally, the workaround is to set Vary before calling stale? / fresh_when:
```
merge_vary_header!("X-Inertia")

fresh_when(...)
```
or:
```
merge_vary_header!("X-Inertia")

if stale?(...)
  render inertia: ...
end
```

Potential upstream improvements:

1. Deduplicate the current renderer Vary merge logic instead of blindly appending X-Inertia.
2. Provide an adapter helper for conditional caching, for example something like inertia_vary!, documented for use before stale? / fresh_when.
3. Or set Vary: X-Inertia earlier in the adapter lifecycle, though that may need care because the controller concern can be included broadly.

The key issue is not that inertia_rails never sets Vary; it does. The issue is that Rails conditional GET can finish the response before the Inertia renderer runs.

----
To be more specific, I have an app, that didn't have that kind of work around. It is something like: open browser, go to the home page, refresh, then navigate a little bit, then click to the home page again, and it displays the legend inertia popup b/c of above problem.

skryukov:
Hey @darkamenosa, sorry for the long pause here.

I pushed a batch of commits while doing the final review pass: added regression specs, polished the docs, moved `xsrf_cookie_refresh` validation into the config getter, and extracted the lazy-refresh logic into a policy object so the concern doesn't inject private methods into app controllers. Behavior is unchanged, suite is green.

Re the Vary / 304 problem: I took a stab at it in #383 by folding the request variant into the ETag instead of touching 304 headers. Thanks for your work on this!

skryukov (approved):

PR #370, opened 2026-04-23
Use params.expect instead of params.require.permit (Rails 8+)

tyrro:
## Summary

Replaces `params.require(:model).permit(...)` with `params.expect(model: [...])` introduced in Rails 8.

`params.expect` is the safer, more explicit successor to `params.require(...).permit(...)`. It raises `ActionController::ParameterMissing` for unexpected input shapes and avoids the [array-of-permitted-scalars bypass](https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-expect) that the old pattern was susceptible to.

Changes:
- **Scaffold generator template** (`lib/generators/inertia/scaffold_controller/templates/controller.rb.tt`): generated controller code now uses `params.expect`
- **Docs** (`docs/guide/validation.md`, `docs/guide/forms.md`, `docs/guide/redirects.md`): updated illustrative examples to reflect the modern Rails 8 API

Reference: https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-expect

## Test plan

- Ran the existing test suite — no failures.
- Verified scaffold output manually: generated controller uses `params.expect(model: [:attr1, :attr2])`.
- Doc changes are illustrative only (no executable code).

## Checklist

- [ ] Tests added or updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+* Use `params.expect` instead of `params.require(...).permit(...)` in scaffold generator and docs
+
 ## [3.21.0] - 2026-04-14
 
 * Skip excluded hash props on partial reloads (@erickreutz)
--- a/docs/guide/forms.md
+++ b/docs/guide/forms.md
@@ -2499,7 +2499,7 @@ class UsersController < ApplicationController
   private
 
   def user_params
-    params.require(:user).permit(:name, :email)
+    params.expect(user: [:name, :email])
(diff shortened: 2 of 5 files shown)


skryukov:
Hey @tyrro! Thanks for the PR! We probably need to check Rails version and conditionally use the new API, since the gem supports Rails 6+ as of today.

tyrro:
Good point, @skryukov! Updated the scaffold generator template to conditionally use `params.expect` on Rails 8+ and fall back to `params.require(...).permit(...)` for Rails 6/7. The docs examples remain as `params.expect` since they're illustrative rather than scaffolded output.

skryukov (approved):

PR #375, opened 2026-05-27
Add CSP nonce to initial page script (when needed)

nicholaspufal:
### What is this?

InertiaRails can render the initial page payload in an `application/json` script tag when `use_script_element_for_initial_page` is enabled. Rails apps with strict CSP settings may require every script tag to include the per-request nonce, but `inertia_root` currently renders that JSON script without one.

In our app, that caused CSP violations during an upgrade and required an app-level monkey patch to add the nonce. This PR moves that behavior into the built-in helper by resolving Rails' `content_security_policy_nonce` when
available and adding it to the script tag.

The existing `data-page` div rendering is preserved when script mode is disabled.

### Example problem

A Rails app using a strict CSP can enable nonces for scripts:

```ruby
# config/initializers/content_security_policy.rb
Rails.application.config.content_security_policy_nonce_generator = ->(_request) { SecureRandom.base64(16) }
Rails.application.config.content_security_policy_nonce_directives = %w[script-src]
```

Then, if the app enables script-based initial page data:

```ruby
  config.use_script_element_for_initial_page = true
end
```

`inertia_root` currently renders:

```html
<script data-page="app" type="application/json">
  {"component":"Users/Index","props":{}}
</script>
<div id="app"></div>
```

That `<script>` has no nonce, so strict CSP setups can reject it even though it is `type="application/json"`.

After this change, when Rails provides a CSP nonce, the rendered HTML becomes:

```html
<script data-page="app" type="application/json" nonce="abc123">
  {"component":"Users/Index","props":{}}
</script>
<div id="app"></div>
```

When `use_script_element_for_initial_page` is disabled, the existing behavior stays the same:

```html
<div id="app" data-page="{...}"></div>
```

### DIFF
--- a/lib/inertia_rails/helper.rb
+++ b/lib/inertia_rails/helper.rb
@@ -36,8 +36,14 @@ module InertiaRails
       id ||= config.root_dom_id
 
       if config.use_script_element_for_initial_page
+        script_options = { 'data-page': id, type: 'application/json' }
+        if respond_to?(:content_security_policy_nonce, true)
+          nonce = content_security_policy_nonce
+          script_options[:nonce] = nonce if nonce.present?
+        end
+
         safe_join([
-                    tag.script(page.to_json.html_safe, 'data-page': id, type: 'application/json'),
+                    tag.script(page.to_json.html_safe, **script_options),
                     tag.div(id: id)
                   ], "\n")
       else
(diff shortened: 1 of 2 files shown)


skryukov (approved):
Thanks for the PR @nicholaspufal (and upvotes from the Doximity gang)!

nicholaspufal:
@skryukov Quick follow up: would it be possible to publish a new release for this?

Let me know if I can help with anything else please. Thank you!

skryukov:
@nicholaspufal released as https://github.com/inertiajs/inertia-rails/releases/tag/v3.21.2 sorry it took a bit to prepare this one 😃

PR #377, opened 2026-06-07
Add `rescue` option to defer props & update docs

skryukov:
## Summary

See inertiajs/inertia-laravel#864.

`InertiaRails.defer(rescue: true) { ... }` catches and reports exceptions raised while resolving a deferred prop instead of failing the whole response. The prop is omitted and its path is returned in a new `rescuedProps` page key so the client can render a fallback (the `rescue` slot/prop on `<Deferred>`).

```ruby
render inertia: {
  permissions: InertiaRails.defer(rescue: true) { current_user.permissions },
}
```

Rails specifics:
- The value is serialized via `as_json` inside the rescued scope, so errors from lazy `ActiveRecord::Relations` / serialization are caught too, not just exceptions raised directly in the block.
- Errors are reported via `Rails.error`, falling back to `Rails.logger` on Rails < 7.

Also syncs assorted guide docs from upstream.

## Checklist

- [x] Tests added or updated

### DIFF
--- a/docs/guide/once-props.md
+++ b/docs/guide/once-props.md
@@ -138,7 +138,7 @@ Returning `nil` on signed-out requests overwrites the remembered user on every r
 ```ruby
 class InertiaController < ApplicationController
   inertia_share(
-    auth: Current.user ? -> { Current.user&.as_json(only: [:id, :name, :email]) } : nil
+    auth: Current.user ? InertiaRails.once { Current.user&.as_json(only: [:id, :name, :email]) } : nil
   )
 end
 ```
--- a/docs/guide/once-props.md
+++ b/docs/guide/once-props.md
@@ -129,6 +129,20 @@ class ApplicationController < ActionController::Base
 end
 ```
 
+## Conditional Once Props
+
+Some shared data is mostly stable but changes at specific moments, like the authenticated user. You may return a once prop while a user is authenticated, and `nil` otherwise. Sign in and sign out usually end with a redirect to a page that does not explicitly request the `auth` prop, so a remembered once value would go stale.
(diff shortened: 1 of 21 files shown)


docs/guide/once-props.md
```diff
@@ -129,6 +129,20 @@ class ApplicationController < ActionController::Base
...
 ```
 
+## Conditional Once Props
+
+Some shared data is mostly stable but changes at specific moments, like the authenticated user. You may return a once prop while a user is authenticated, and `nil` otherwise. Sign in and sign out usually end with a redirect to a page that does not explicitly request the `auth` prop, so a remembered once value would go stale.
+
+Returning `nil` on signed-out requests overwrites the remembered user on every response, while returning a once prop when authenticated lets the client remember the user across visits.
+
+```ruby
+class InertiaController < ApplicationController
```
bknoles, line 139: Should this be something like this instead?

```ruby
inertia_share(
    auth: Current.user ? -> {
      InertiaRails.once{ Current.user&.as_json(only: [:id, :name, :email]) }
    } : nil
  )
```

docs/guide/once-props.md
```diff
@@ -129,6 +129,20 @@ class ApplicationController < ActionController::Base
...
+  inertia_share(
+    auth: Current.user ? -> { Current.user&.as_json(only: [:id, :name, :email]) } : nil
```
skryukov:
```suggestion
    auth: Current.user ? InertiaRails.once { Current.user&.as_json(only: [:id, :name, :email]) } : nil
```

bknoles (approved):

PR #379, opened 2026-06-11
Fix external redirect handling for Inertia requests

skryukov:
Redirecting an Inertia app to another origin (OAuth, payment providers) cannot work with a plain redirect: the browser follows XHR redirects transparently, and the follow-up request to the external origin fails CORS checks. The only mechanism that works is Inertia's location response (409 + `X-Inertia-Location`), which triggers a full `window.location` visit. This PR makes that path automatic where possible and ergonomic where it cannot be.

### 1. The middleware converts cross-origin redirects automatically

For Inertia requests, a 301/302/303 whose `Location` points to another origin (scheme, host, or port differs) becomes a 409 + `X-Inertia-Location`.

- The response is rewritten in place, so `Set-Cookie` and all other headers survive (matters mid-OAuth).
- 307/308 are left untouched: a full page visit cannot preserve the HTTP method.
- Pending Inertia session options are cleared, matching `inertia_location` semantics.
- Opt out globally (`config.convert_external_redirects = false`), per controller (`inertia_config`), or via the `INERTIA_CONVERT_EXTERNAL_REDIRECTS` env var.

inertia-laravel has no equivalent (its middleware only rewrites fragment-bearing redirects into 409s); this part is novel to inertia-rails.

### 2. `redirect_to url, inertia: { full_page: true }` for same-origin targets

Same-origin redirects to non-Inertia endpoints fail differently: the XHR follows the redirect, the target responds without the `X-Inertia` header, and the client rejects it with the "invalid response" modal. These cannot be converted automatically, because a `Location` header does not reveal whether its target renders an Inertia page. The new option marks them explicitly:

- Think of it as `inertia_location` with `redirect_to` ergonomics: the controller lets `redirect_to` run, then rewrites the response in place (`Location` becomes `X-Inertia-Location`, 409, empty body). URL helpers, records, flash options, and cookies behave as usual.
- Plain clients still get the regular redirect.
- Converts independently of `convert_external_redirects`; combining it with a method-preserving 307/308 raises `ArgumentError`.

### 3. `inertia_location` handles plain clients

A non-Inertia request (a plain form POST, a direct browser visit) previously got a bare 409 dead end. It now gets a regular redirect, mirroring inertia-laravel's `ResponseFactory::location()`.

> [!WARNING]
> The plain-client branch passes `allow_other_host: true`, bypassing Rails' open redirect protection. Only trusted URLs should ever reach `inertia_location`. The docs carry the same warning.

### Also fixed

A stale asset version no longer replaces a location response with a forced refresh to the current URL: a location response is already a full page visit, so there is nothing to refresh. This also fixes a pre-existing quirk where `inertia_location` on a stale GET was silently overridden by the refresh.

## Deliberate non-features

Both verified by probing what the middleware actually receives at its stack position (innermost, directly below `Rack::ETag`) on Rack 2 and Rack 3:

- **No header-casing indirection.** Rails-originated responses carry a case-insensitive `Rack::Headers` (Rack 3) or a capitalized plain hash (Rack 2), so capitalized header literals work on both. Redirects from hand-written Rack endpoints pass through unconverted, which is the pre-PR behavior.
- **No `Vary: X-Inertia` on redirects.** `Rack::ETag` stamps `Cache-Control: no-cache` on every redirect leaving a default Rails stack, and neither 409 nor 302/303 is heuristically cacheable, so a shared cache never reuses either variant of a converted redirect. Apps that explicitly opt redirects into public caching are responsible for `Vary` themselves.

## Docs

The external-redirects guide is restructured around the split: cross-origin (automatic) vs same-origin (explicit `full_page`), with `inertia_location` documented as the manual option for older versions, plus the open-redirect warning above.

## Test plan

`bundle exec rspec spec/inertia`: 851 examples green on Rails 8.1 / Rack 3 locally, full matrix in CI. Notable cases: controller- and route-level conversion for 301/302/303; 307/308 passthrough; same-origin, relative, scheme-relative, invalid-URI, default-port, and host-casing negatives; `Set-Cookie` preservation; stale-version precedence; cookie and flash survival through `full_page`; session-option clearing; global and per-controller opt-out; plain-client passthrough.

## Checklist

- [x] Tests added or updated


docs/guide/redirects.md
```diff
@@ -68,10 +68,63 @@ redirect_to article_path(article), inertia: { preserve_fragment: true }
...
 ```ruby
-inertia_location index_path
+inertia_location 'https://checkout.stripe.com/session_123'
 ```
 
-The `inertia_location` method will generate a `409 Conflict` response and include the destination URL in the `X-Inertia-Location` header. When this response is received client-side, Inertia will automatically perform a `window.location = url` visit.
+For Inertia requests, it generates a `409 Conflict` response with the destination URL in the `X-Inertia-Location` header. For non-Inertia requests (for example, a direct browser visit to the same endpoint), it performs a regular redirect — on older versions of Inertia Rails, it responds with `409 Conflict` regardless of the request type.
+
+> [!WARNING]
+> `inertia_location` bypasses Rails' open redirect protection. Only pass trusted URLs to it, and don't redirect to user-provided URLs.
```
bknoles, line 130: Is this new behavior? Or does that match the existing behavior?

skryukov: match existing one

bknoles (approved):

PR #382, opened 2026-06-25
Add @inertiajs/core as a direct dependency for TypeScript installs

### LINKED ISSUE #380: TypeScript compiler errors on freshly generated scaffold when using pnpm
### Describe the bug

The generated scaffolding files have TypeScript compilation mismatches. While `npm run check` passes successfully without any issues, running `pnpm check` fails with multiple type errors.

This issue specifically occurs under a `pnpm` environment, likely due to pnpm's strict dependency resolution and module structure.

The expected behavior is for `pnpm check` to pass successfully even when using pnpm.

### Steps to reproduce

Create a new Rails app and install Inertia with React/TypeScript/Vite via pnpm.
Generate a scaffold and run type-check.

```bash
rails new my_app --skip-js
cd my_app
bundle add inertia_rails
rails generate inertia:install --package-manager=pnpm --vite --typescript --framework=react --tailwind --force --verbose
rails g inertia:scaffold User name
rails db:migrate
pnpm check
```

### Environment

- inertia_rails version: 3.21.2
- Rails version: 8.1.3
- Ruby version: ruby 4.0.5
- Frontend adapter and version (e.g. @inertiajs/react 3.0): 3.4.0
- Build tool (e.g. Vite Ruby, RailsVite, etc): 3.10.2

### Additional context

```shell
% pnpm check
$ tsc -p tsconfig.app.json && tsc -p tsconfig.node.json
app/frontend/pages/users/form.tsx:1:41 - error TS2307: Cannot find module '@inertiajs/core' or its corresponding type declarations.

1 import { type FormComponentProps } from '@inertiajs/core'
                                          ~~~~~~~~~~~~~~~~~

app/frontend/pages/users/form.tsx:31:30 - error TS2339: Property 'join' does not exist on type 'string'.

31                 {errors.name.join(', ')}
                                ~~~~

app/frontend/pages/users/index.tsx:17:9 - error TS2322: Type 'unknown' is not assignable to type 'ReactNode'.

 17         {flash.notice && (
            ~~~~~~~~~~~~~~~~~~
 18           <p className="py-2 px-3 bg-green-50 mb-5 text-green-500 font-medium rounded-lg inline-block">
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
...
 20           </p>
    ~~~~~~~~~~~~~~
 21         )}
    ~~~~~~~~~~

app/frontend/pages/users/index.tsx:19:13 - error TS2322: Type '{}' is not assignable to type 'ReactNode'.

19             {flash.notice}
               ~~~~~~~~~~~~~~

  node_modules/.pnpm/@types+react@19.2.17/node_modules/@types/react/index.d.ts:2267:9
    2267         children?: ReactNode | undefined;
                 ~~~~~~~~
    The expected type comes from property 'children' which is declared here on type 'DetailedHTMLProps<HTMLAttributes<HTMLParagraphElement>, HTMLParagraphElement>'

app/frontend/pages/users/show.tsx:18:11 - error TS2322: Type 'unknown' is not assignable to type 'ReactNode'.

 18           {flash.notice && (
              ~~~~~~~~~~~~~~~~~~
 19             <p className="py-2 px-3 bg-green-50 mb-5 text-green-500 font-medium rounded-lg inline-block">
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
...
 21             </p>
    ~~~~~~~~~~~~~~~~
 22           )}
    ~~~~~~~~~~~~

app/frontend/pages/users/show.tsx:20:15 - error TS2322: Type '{}' is not assignable to type 'ReactNode'.

20               {flash.notice}
                 ~~~~~~~~~~~~~~

  node_modules/.pnpm/@types+react@19.2.17/node_modules/@types/react/index.d.ts:2267:9
    2267         children?: ReactNode | undefined;
                 ~~~~~~~~
    The expected type comes from property 'children' which is declared here on type 'DetailedHTMLProps<HTMLAttributes<HTMLParagraphElement>, HTMLParagraphElement>'


Found 6 errors in 3 files.

Errors  Files
     2  app/frontend/pages/users/form.tsx:1
     2  app/frontend/pages/users/index.tsx:17
     2  app/frontend/pages/users/show.tsx:18
[ELIFECYCLE] Command failed with exit code 1.
```

heyrobertchang:
Fixes #380.

On a fresh React + TypeScript app installed with pnpm, `pnpm check` fails with 6 TypeScript errors; the same app passes with npm.

## Root cause

The generator ships `types/globals.d.ts`, which augments `@inertiajs/core` to type shared props, flash data, and form errors:

```ts
declare module '@inertiajs/core' {
  export interface InertiaConfig {
    sharedPageProps: SharedProps
    flashDataType: FlashData
    errorValueType: string[]
  }
}
```

Module augmentation only applies if TypeScript can resolve `@inertiajs/core` from the augmenting file. The package was never declared in package.json — it's only a transitive dependency of `@inertiajs/react`. npm hoists it to the top of node_modules, so resolution works by accident. pnpm's strict layout keeps it inside `.pnpm`, so resolution fails, the augmentation is silently ignored, and flash/error types revert to their defaults. All 6 errors trace back to that single miss: the failing `import { type FormComponentProps } from '@inertiajs/core'`, `errors.name.join` (errors typed `string` instead of `string[]`), and `flash.notice` not being a valid `ReactNode`.

The scaffold templates are untouched — `.join` is correct once `errorValueType: string[]` applies, and removing it would break npm setups.

## Fix

Declare `@inertiajs/core` as a devDependency when TypeScript is enabled, pinned to the same `--inertia-version` as the adapter (default `latest`). The adapter pins core exactly (`@inertiajs/react@3.5.0` depends on `@inertiajs/core@3.5.0`), so both resolve to one deduped copy. An explicit `--inertia-version=3.4.0` lands as an exact pin for both, with zero room to drift. With the default `latest`, both land as caret ranges; in theory a stale adapter lockfile plus a newer core minor could yield two copies, but core and the adapters ship in lockstep, so the ranges resolve identically in practice.

## Testing

- Reproduced #380 on a fresh React/TS app with pnpm on Rails 8.1: same 6 errors, no `@inertiajs/core` at the top of node_modules.
- After the fix: `pnpm check` passes, core appears at the top of node_modules at the adapter's version, and `npm run check` still passes.
- With `--inertia-version=3.4.0`: core lands as an exact `3.4.0` matching the adapter, one copy installed, check passes.
- Full RSpec suite green (790 examples, 0 failures).

### DIFF
--- a/lib/generators/inertia/install/install_generator.rb
+++ b/lib/generators/inertia/install/install_generator.rb
@@ -142,13 +142,8 @@ module Inertia
       def install_typescript
         say 'Adding TypeScript support'
 
-        # @inertiajs/core is only a transitive dependency of the framework
-        # adapters, so it isn't exposed at the top level under pnpm's strict
-        # node_modules layout. The generated `types/globals.d.ts` augments the
-        # `@inertiajs/core` module (to type flash/shared props and errors), and
-        # the React/Svelte scaffolds import types from it directly. Declaring it
-        # as an explicit dependency keeps it resolvable across package managers.
-        # Pin it to the same version as the adapter so the types stay in sync.
+        # globals.d.ts augments @inertiajs/core and scaffolds import its types;
+        # pnpm won't expose the adapters' transitive dep, so declare it directly.
         add_dependencies(
           "@inertiajs/core@#{options[:inertia_version]}",
           *FRAMEWORKS[framework]['packages_ts'],
--- a/lib/generators/inertia/install/install_generator.rb
+++ b/lib/generators/inertia/install/install_generator.rb
@@ -142,7 +142,18 @@ module Inertia
       def install_typescript
         say 'Adding TypeScript support'
 
-        add_dependencies(*FRAMEWORKS[framework]['packages_ts'], dev: true)
+        # @inertiajs/core is only a transitive dependency of the framework
+        # adapters, so it isn't exposed at the top level under pnpm's strict
+        # node_modules layout. The generated `types/globals.d.ts` augments the
+        # `@inertiajs/core` module (to type flash/shared props and errors), and
+        # the React/Svelte scaffolds import types from it directly. Declaring it
+        # as an explicit dependency keeps it resolvable across package managers.
+        # Pin it to the same version as the adapter so the types stay in sync.
+        add_dependencies(
+          "@inertiajs/core@#{options[:inertia_version]}",
+          *FRAMEWORKS[framework]['packages_ts'],
+          dev: true
+        )
 
         say 'Copying tsconfig and types'
 


skryukov (approved):

skryukov:
Thanks @heyrobertchang!

PR #384, opened 2026-07-07
Adopt Inertia.js v3.5 serverHead for server driven meta tags

skryukov:
## Summary

Adopts Inertia.js v3.5's [`serverHead` option](https://github.com/inertiajs/inertia/pull/3161) as the client side of [server driven meta tags](https://inertia-rails.dev/cookbook/server-managed-meta-tags), replacing the copy-pasted cookbook client component while keeping the whole `inertia_meta` backend UX, automatic head-key generation, and the crawler-friendly no-SSR initial HTML (which `serverHead` alone does not provide).

```ruby
# config/initializers/inertia_rails.rb
config.server_head = true
```

```js
createInertiaApp({
  serverHead: true,
})
```

With `config.server_head = true`, the renderer serializes meta tags into the `head` prop as ready-to-render HTML strings — the prop name the client's zero-config `serverHead: true` reads. The `inertia_meta_tags` layout helper emits byte-identical `data-inertia`-keyed strings, so crawlers see the tags without SSR and hydration adopts them without touching the DOM (`isEqualNode`).

### Design notes

- `config.server_head` accepts `true` or a String (custom prop name) for apps that already use a `head` prop; ENV `INERTIA_SERVER_HEAD` supports both forms; per-controller via `inertia_config`. Boolean-ish ENV strings (`TRUE`, `1`, `""`, `off`) normalize instead of silently becoming prop names nothing reads.
- The target prop name is **reserved** while `server_head` is enabled: the renderer raises `InertiaRails::Error` whenever a page defines a prop with that name, instead of silently clobbering it. The check runs against the pre-resolution props (symbolized, shared data included), so partial reloads, deferred/optional props, and string-keyed `prop_transformer` output cannot make the raise intermittent — without this, an app `head` prop slips through on exactly the request shapes developers don't watch, and either gets silently replaced on the client or crashes `inertia_meta_tags` on pages without meta tags. The message points at the custom-key escape hatch.
- `Configuration#head_attribute` is the single source of truth for the `data-inertia`/`inertia` marker shared by renderer and helper; `server_head` forces `data-inertia` (the only attribute v3 core recognizes).
- Strings in the meta prop that did not come from the renderer are escaped by `safe_join` (previously they crashed the helper).
- The structured `_inertia_meta` format and the cookbook component keep working unchanged for older clients; the plan is to deprecate them in 4.0.

### Also included (separate commits)

- Fix `MetaTag#to_tag` reading the *global* config, which ignored per-controller `use_data_inertia_head_attribute` set via `inertia_config`.
- SHA256 instead of MD5 for auto-generated head-key digests — MD5 raises on FIPS-enabled Rubies. The generated keys are opaque per-response values, so the change is invisible.
- Remove the unused controller argument from `MetaTagBuilder.new`.
- Install generator preconfigures `serverHead: true` (all six entrypoint templates) + `config.server_head = true`, so new apps get server driven meta tags out of the box.

## Test plan

- [x] 837 examples, 0 failures; RuboCop clean; docs build passes. Every guard branch has a regression spec verified to fail against the pre-fix code (blank-meta collision, deferred prop on initial load, partial reload excluding the conflicting prop, string-keyed transformer, ENV normalization).
- [x] Wire format spec-locked: exact strings in `head`, custom prop name, collision raise, JSON-payload escape (no `</script>` breakout), full-stack layout render via a dummy layout using `inertia_meta_tags`.
- [x] Protocol-tested against the real `@inertiajs/core` 3.5.0 head manager (keys preserved, page `<Head>` overrides by key, absent prop clears server tags).
- [x] Browser-verified with `@inertiajs/react` 3.6.1 on a freshly generated scaffold (also proving the generator templates): zero-mutation hydration adoption, title/description/JSON-LD swap on client-side visits, zero-mutation partial reloads (DOM element identity probe-verified), `history.back()` restore from encrypted history, page-level `<Head>` override of server tags, absent-prop tag clearing, clean console.
- [x] Docs followability: an agent running a weak model set the feature up on a fresh app using only the cookbook + configuration pages, and completed every step without blocking confusion. The two gaps it flagged (whether `inertia_ssr_head` can coexist without SSR, and the missing-default-title footgun confirmed in the browser) are now addressed in the docs.
- [x] SSR path (`serverHead` in the SSR entrypoint feeding `inertia_ssr_head`) — from upstream implementation reading, not exercised locally.


bknoles (approved):
Love how few changes we needed to support this! So much better than copy-pasta-ing a component

PR #385, opened 2026-07-07
Fix generator template endpoint deployment on Cloudflare Pages

skryukov:
## Summary

Makes `rails new myapp -m https://inertia-rails.dev` actually work in production.

The Pages Function added in #365 (`docs/functions/index.ts`) never deployed: the docs build artifact only contains `docs/.vitepress/dist`, and the deploy workflow publishes that directory without checking out the repo — so wrangler never sees the `functions/` directory and silently ships static assets only. Verified live: `curl -H "Accept: application/x-thor-template" https://inertia-rails.dev/` currently returns the docs HTML instead of the template.

This PR replaces the function with an advanced-mode `_worker.js` in `docs/public/`. VitePress copies `public/` files into the build output verbatim, so the worker travels **inside the artifact** and deploys with zero workflow changes — including for PR preview deployments, which is what lets us verify this exact code path before merging.

`_routes.json` restricts worker invocations to `/` and `/template.rb`; all other docs traffic is served statically as before (no worker latency/quota impact).

Behavior:
- `GET /` with Thor's `Accept: application/x-thor-template` → proxies the compiled template from `inertia-rails/generator@dist`
- `GET /template.rb` → same template, unconditionally (inspectable in a browser, usable with `rails app:template`)
- everything else → static docs site, unchanged

## Test plan

- [x] Local `vitepress build` puts `_worker.js` and `_routes.json` at the dist root
- [x] `wrangler pages dev` locally: Thor Accept header → template; `/template.rb` → template; no header → docs HTML; static assets unaffected
- [x] Against the Cloudflare Pages **preview deployment** of this PR: `curl -H "Accept: application/x-thor-template" <preview-url>` returns the template, plain curl returns docs HTML
- [x] Full end-to-end: `rails new testapp -m <preview-url>` completes using the template (see comment below)

## Checklist

- [ ] Tests added or updated
- [ ] `CHANGELOG.md` updated (for user-facing changes)


skryukov:
Verified against the preview deployment (`https://inertiajs-docs-template-endp.inertia-rails.pages.dev`):

- `curl -H "Accept: application/x-thor-template" <preview>/` → returns the Ruby template with `Content-Type: application/x-thor-template`
- `curl <preview>/template.rb` → returns the template
- plain `curl <preview>/` → docs HTML, static assets unaffected
- Full end-to-end: `rails new testapp -m https://inertiajs-docs-template-endp.inertia-rails.pages.dev` (Rails 8.1.3) fetched the template via content negotiation and completed with "Inertia Rails installed successfully!" — generated app has `inertia_rails ~> 3.21`, `@inertiajs/react 3.6.1`, and the initializer in place.

Since preview and production deploy the identical artifact (worker included), merging this makes `rails new myapp -m https://inertia-rails.dev` work in production.

bknoles:
Is this going to generate serverless function costs every time someone uses the install generator?

skryukov:
Free plan comes with 100,000 daily requests. But for all requests, not only installs.

PR #389, opened 2026-07-13
Add static props support to the inertia routes helper

skryukov:
## Summary

Adds Laravel `Route::inertia` parity to the `inertia` routes helper:

```ruby
inertia 'about' => 'About', props: { title: 'About us' }
```

The hash is stored in the route defaults and handed to the Inertia render by `StaticController`, so static pages can carry static props without a dedicated controller.

## Design notes

- The props hash lives in the route defaults and is therefore shared across requests. `StaticController` passes a `deep_dup` to the render: `expand_dot_notation` in the props resolver can write into nested input hashes (a plain `user: {...}` pair is stored by reference and a later dotted `'user.name'` key mutates it), and a shared hash would persist that mutation across requests.
- Docs updated (`guide/routing.md`) with an `@available_since rails=master` marker.

Stacked on #388 (the option-handling fix); the base branch is set accordingly and this PR should be rebased or retargeted once #388 merges.

## Test plan

- `bundle exec rspec spec/inertia` — 791 examples green.
- New spec asserts the props arrive in the Inertia page JSON.

## Checklist

- [x] Tests added or updated


bknoles (approved):
Neat!

PR #395, opened 2026-07-17
Only set the XSRF-TOKEN cookie on HTML and XHR responses

### LINKED ISSUE #394: Global XSRF-TOKEN after_action sets Set-Cookie on every response, making CDN caching impossible (e.g. ActiveStorage image proxies)
### Versions

- `inertia_rails` 3.21.2
- Rails 8.1.3

### Summary

`InertiaRails::Controller` is included into `ActionController::Base` itself:

```ruby
# lib/inertia_rails/engine.rb
initializer 'inertia_rails.action_controller' do
  ActiveSupport.on_load(:action_controller_base) do
  end
end
```

so this after_action runs on **every response from every controller**, not just Inertia ones:

```ruby
# lib/inertia_rails/controller.rb
after_action do
  cookies['XSRF-TOKEN'] = form_authenticity_token if protect_against_forgery?
end
```

I understand this is deliberate — #71 explicitly wanted non-Inertia requests to seed the token so CSRF works out of the box, and that part works well. The side effect I'd like advice on is that it also applies to responses that aren't HTML and aren't consumed by axios at all — in particular **ActiveStorage's image controllers**, which also inherit from `ActionController::Base`.

**Cloudflare (and most CDNs) will not cache a response carrying `Set-Cookie`.** So every image we serve through our CDN is bypassed, despite ActiveStorage marking it `public, immutable, max-age=100y`.

### Reproduction

Any Rails app that puts `ActiveStorage::Representations::ProxyController` / `Blobs::ProxyController` behind a CDN. A request for an image variant returns:

```
content-type:      image/png
cache-control:     max-age=3155695200, public, immutable
set-cookie:        XSRF-TOKEN=<token>; path=/; samesite=lax; secure
cf-cache-status:   BYPASS
```

A `Set-Cookie` on a PNG. Minimal request-spec repro (forgery protection on, as in production):

```ruby
blob = ActiveStorage::Blob.create_and_upload!(
  io: StringIO.new("x"), filename: "a.png", content_type: "image/png"
)
get rails_service_blob_proxy_path(blob.signed_id, blob.filename)

response.headers["Set-Cookie"]
# => "XSRF-TOKEN=...; path=/; samesite=lax"
```

### Evidence that Set-Cookie is the cause

Two requests to the **same host**, same page, same `vary: Origin, accept-encoding`:

| | ActiveStorage proxy (Rails) | static asset (not Rails) |
|---|---|---|
| `cache-control` | `max-age=3155695200, public, immutable` | `public, max-age=14400` |
| `set-cookie` | `XSRF-TOKEN=…` | none |
| `cf-cache-status` | **BYPASS** | **HIT** (`age: 2178`) |

The static file asks for 4 hours and is cached; the image asks for 100 years and is not. In a HAR from a real page load, 32/33 image requests were `BYPASS` (all of them representation proxies) and stayed that way across refreshes.

### Note: `skip_forgery_protection` does not help

The natural workaround doesn't work, which may be worth documenting either way. `skip_forgery_protection` only removes the `verify_authenticity_token` callback; it leaves `protect_against_forgery?` **true**, which is the flag this after_action checks. `ActiveStorage::DiskController` calls `skip_forgery_protection` in Rails itself and still gets the cookie.

The only lever we found is clearing `allow_forgery_protection`, which is a bigger hammer than we'd like:

```ruby
# our current workaround
Rails.application.config.to_prepare do
  [
    ActiveStorage::Blobs::ProxyController,
    ActiveStorage::Blobs::RedirectController,
    ActiveStorage::Representations::ProxyController,
    ActiveStorage::Representations::RedirectController
  ].each { |c| c.allow_forgery_protection = false }
end
```

That works (verified: `Set-Cookie` gone), but it conflates "don't verify CSRF here" with "don't advertise a token here", and it needs re-applying for any other engine that mounts controllers under `ActionController::Base`.

### Question / possible directions

What would you consider the right fix? Some options, happy to open a PR for whichever you prefer:

1. **Only set the cookie on HTML responses** — e.g. guard on `response.media_type == "text/html"`. Non-HTML responses can't be the thing that seeds a token for axios anyway, so this seems to preserve #71's intent while excluding images/streams.
2. **A config flag** — `InertiaRails.configure { |c| c.xsrf_cookie = false }`, plus perhaps a per-controller opt-out.
3. **Skip when the response is already committed / streaming** (ActiveStorage streams its bodies).
4. **Don't include the concern globally** — only into controllers that actually render Inertia responses. Presumably rejected in #71, since the token has to be seeded from ordinary pages too.

(1) feels like the smallest change that keeps out-of-the-box CSRF working, but I don't know what other consumers rely on the cookie appearing on non-HTML responses — hence asking rather than sending a PR straight away.

Thanks for the library!

acetinick:
Fixes #394. Implements option 1 from that thread, as @skryukov suggested — including the correction that the guard belongs on the **request**, not the response.

### The problem

`InertiaRails::Controller` is included into `ActionController::Base`, so the `XSRF-TOKEN` after_action runs for every response from every controller — including ones that never render Inertia and whose clients never read the cookie.

ActiveStorage is the clearest case: its image controllers inherit from `ActionController::Base` too, so a **PNG** comes back with `Set-Cookie: XSRF-TOKEN`. CDNs bypass their cache for any response carrying `Set-Cookie`, so images ActiveStorage marks `public, immutable, max-age=100y` are never edge-cached.

Measured on our production CDN — two requests to the same host, same page:

| | ActiveStorage image proxy | static asset (not Rails) |
|---|---|---|
| `vary` | `Origin, accept-encoding` | identical |
| `cache-control` | `max-age=3155695200, public, immutable` | `public, max-age=14400` |
| `set-cookie` | `XSRF-TOKEN=…` | none |
| `cf-cache-status` | **BYPASS** | **HIT** (`age: 2178`) |

The static file asks for 4 hours and is cached; the image asks for 100 years and is not. In a HAR from one real page load, 32 of 33 image requests were `BYPASS`.

### The change

```ruby
next unless request.format.html? || request.xhr?
```

Only page loads and XHR (Inertia visits, `useHttp`) can read the cookie and send it back, so nothing else needs it.

Guarding on `response.media_type` — my original suggestion in the issue — would break conditional responses, since a `304 Not Modified` carries no `Content-Type`. The request-side guard doesn't have that problem, and the existing spec for it (`rewrites the XSRF-TOKEN cookie even on 304 Not Modified responses by default`) still passes unchanged.

### Tests

Two specs added alongside the existing ones in `spec/inertia/request_spec.rb`:

- an XHR call that isn't HTML → still gets the cookie
- a request that is neither HTML nor XHR (`Accept: image/png`) → doesn't

Verified the second one fails without the guard:

```
Failure/Error: it { is_expected.not_to include('XSRF-TOKEN') }
  expected {"XSRF-TOKEN" => "tiEoVzYPyNU_9l-40JkdT28YvO-..."} not to include "XSRF-TOKEN"
```

`spec/inertia` is green locally (58 examples, 0 failures on Ruby 3.4). Leaving the full matrix to CI.

### Notes

- This composes with `xsrf_cookie_refresh: :lazy` (#368) rather than replacing it — the new guard runs first, the policy still decides refresh behaviour for HTML/XHR.
- Happy to adjust the wording of the comment or the spec placement if you'd prefer them elsewhere.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+* Only set the `XSRF-TOKEN` cookie on HTML and XHR responses, so responses no client reads it from — notably ActiveStorage's image endpoints, which also inherit from `ActionController::Base` — no longer carry a `Set-Cookie` that stops CDNs from caching them (@acetinick)
 * Add `xsrf_cookie_refresh` configuration option. Set to `:lazy` to skip rewriting the `XSRF-TOKEN` cookie on safe requests when a valid cookie is already present, keeping conditionally cached (`ETag`/`304`) responses free of `Set-Cookie` churn (@darkamenosa)
 * Fix `inertia: { errors: }` and `inertia: { clear_history: }` being discarded when `redirect_to` is given an explicit non-302 status such as `status: :see_other` (303/307/308) (@skryukov)
 * Fix cross-variant `304 Not Modified` responses under HTTP conditional caching (`fresh_when` / `stale?`) by folding the Inertia request headers into the ETag, so HTML, JSON, and partial-reload representations of the same URL no longer share a validator (@skryukov)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-* Only set the `XSRF-TOKEN` cookie on HTML and XHR responses, so responses no client reads it from — notably ActiveStorage's image endpoints, which also inherit from `ActionController::Base` — no longer carry a `Set-Cookie` that stops CDNs from caching them (@acetinick)
+* Only set the `XSRF-TOKEN` cookie on HTML and XHR responses. Other responses no longer carry a `Set-Cookie` that keeps CDNs from caching them — notably ActiveStorage's images, since its controllers also inherit from `ActionController::Base` (@acetinick)
 * Add `xsrf_cookie_refresh` configuration option. Set to `:lazy` to skip rewriting the `XSRF-TOKEN` cookie on safe requests when a valid cookie is already present, keeping conditionally cached (`ETag`/`304`) responses free of `Set-Cookie` churn (@darkamenosa)
 * Fix `inertia: { errors: }` and `inertia: { clear_history: }` being discarded when `redirect_to` is given an explicit non-302 status such as `status: :see_other` (303/307/308) (@skryukov)
 * Fix cross-variant `304 Not Modified` responses under HTTP conditional caching (`fresh_when` / `stale?`) by folding the Inertia request headers into the ETag, so HTML, JSON, and partial-reload representations of the same URL no longer share a validator (@skryukov)
--- a/lib/inertia_rails/controller.rb
+++ b/lib/inertia_rails/controller.rb
@@ -17,11 +17,8 @@ module InertiaRails
 
       after_action do
         next unless protect_against_forgery?
-        # Only clients that can read the cookie and send it back need it: page
-        # loads and XHR (Inertia visits, `useHttp`). Setting it on anything else
-        # puts a `Set-Cookie` on responses that never asked for one — notably
-        # ActiveStorage's image controllers, which also inherit from
-        # ActionController::Base, and which CDNs then refuse to cache.
+        # Included into ActionController::Base, so this runs for non-Inertia
+        # responses too — ActiveStorage images, where it only breaks CDN caching.
         next unless request.format.html? || request.xhr?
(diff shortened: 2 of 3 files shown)


skryukov (approved):
Awesome, thanks @acetinick!

acetinick:
Heads-up on CI: the run here is sitting at `action_required` (first-time contributor), so it needs an approval before it'll go.

In the meantime I ran the identical workflow on my fork — **19/19 jobs green, 820 examples, 0 failures**, RuboCop included, across the full Ruby x Rails 6.1-8.1 matrix: https://github.com/acetinick/inertia-rails/actions/runs/29571887931

Notably `rewrites the XSRF-TOKEN cookie even on 304 Not Modified responses by default` passes unchanged — the case you flagged where `response.media_type` would have broken.

acetinick:
@skryukov amazing to have this merged! do you think we can cut a release on this? would love to test this in our staging

skryukov:
@acetinick sure, just released it as `3.22.0`

PR #397, opened 2026-07-17
Ignore stray X-Inertia header on non-Inertia controllers

### LINKED ISSUE #393: X-Inertia header on an ActionController::API endpoint raises NoMethodError (500)
## Summary

A request carrying an `X-Inertia` header to a controller that isn't an Inertia controller raises `NoMethodError: undefined method 'inertia_configuration'` inside the middleware, which comes back as a 500. The header alone is enough; no credentials or session needed.

This affects any app with the gem in its Gemfile, whether or not it uses Inertia. The engine installs the middleware unconditionally, but only mixes `Controller` into `ActionController::Base`.

Verified on `master` @ [`e2d3881`](https://github.com/inertiajs/inertia-rails/commit/e2d3881) and on the released [v3.21.2](https://rubygems.org/gems/inertia_rails/versions/3.21.2).

## Reproduction (against your own `spec/dummy`)

[`spec/dummy`](https://github.com/inertiajs/inertia-rails/tree/e2d3881/spec/dummy) has no `ActionController::API` controller, so reproducing this takes two small additions:

```ruby
# spec/dummy/app/controllers/inertia_api_test_controller.rb
class InertiaApiTestController < ActionController::API
  def index
    render json: { ok: true }
  end
end
```

```ruby
# spec/dummy/config/routes.rb
get 'api_test' => 'inertia_api_test#index'
```

```ruby
# spec/inertia/api_controller_spec.rb
RSpec.describe 'X-Inertia against an ActionController::API controller', type: :request do
  it 'serves a normal request' do
    get api_test_path
    expect(response.status).to eq 200          # passes
  end

  it 'raises when the X-Inertia header is present' do
    get api_test_path, headers: { 'X-Inertia' => 'true' }
    expect(response.status).to eq 200          # fails: NoMethodError
  end
end
```

```
NoMethodError:
  undefined method 'inertia_configuration' for an instance of InertiaApiTestController
# ./lib/inertia_rails/middleware.rb:103:in 'server_version'
# ./lib/inertia_rails/middleware.rb:107:in 'coerce_version'
# ./lib/inertia_rails/middleware.rb:99:in  'version_stale?'
# ./lib/inertia_rails/middleware.rb:51:in  'stale_inertia_request?'
# ./lib/inertia_rails/middleware.rb:47:in  'keep_inertia_session_options?'
# ./lib/inertia_rails/middleware.rb:33:in  'response'
# ./lib/inertia_rails/middleware.rb:10:in  'call'
```

That path on `master` @ `e2d3881`, frame by frame: [`call`](https://github.com/inertiajs/inertia-rails/blob/e2d3881/lib/inertia_rails/middleware.rb#L9-L12) → [`response`](https://github.com/inertiajs/inertia-rails/blob/e2d3881/lib/inertia_rails/middleware.rb#L19-L42) → [`keep_inertia_session_options?`](https://github.com/inertiajs/inertia-rails/blob/e2d3881/lib/inertia_rails/middleware.rb#L46-L48) → [`stale_inertia_request?`](https://github.com/inertiajs/inertia-rails/blob/e2d3881/lib/inertia_rails/middleware.rb#L50-L52) → [`version_stale?`](https://github.com/inertiajs/inertia-rails/blob/e2d3881/lib/inertia_rails/middleware.rb#L98-L100) → [`coerce_version`](https://github.com/inertiajs/inertia-rails/blob/e2d3881/lib/inertia_rails/middleware.rb#L106-L108) → [`server_version`](https://github.com/inertiajs/inertia-rails/blob/e2d3881/lib/inertia_rails/middleware.rb#L102-L104).

A few things follow from it:

- It isn't GET-only. The reach is [`#response`](https://github.com/inertiajs/inertia-rails/blob/e2d3881/lib/inertia_rails/middleware.rb#L33) → [`#keep_inertia_session_options?`](https://github.com/inertiajs/inertia-rails/blob/e2d3881/lib/inertia_rails/middleware.rb#L46-L48), which has no method gate, so POST raises the same way.
- `X-Inertia-Version` isn't required. [`coerce_version`](https://github.com/inertiajs/inertia-rails/blob/e2d3881/lib/inertia_rails/middleware.rb#L106-L108) asks for `server_version` before it compares anything (`server_version.is_a?(Numeric) ? …`), so `X-Inertia: true` on its own reaches the crash.
- It isn't limited to error responses. The middleware wraps the app, so [`@app.call(@env)`](https://github.com/inertiajs/inertia-rails/blob/e2d3881/lib/inertia_rails/middleware.rb#L21-L25) completes normally and the raise happens on the way back out during post-processing. An endpoint returning a healthy 200 raises too.

## Cause

[`middleware.rb:94-96`](https://github.com/inertiajs/inertia-rails/blob/v3.21.2/lib/inertia_rails/middleware.rb#L94-L96) on v3.21.2 ([`:102-104`](https://github.com/inertiajs/inertia-rails/blob/e2d3881/lib/inertia_rails/middleware.rb#L102-L104) on `master`):

```ruby
def server_version
  (controller&.send(:inertia_configuration) || InertiaRails.configuration).version
end
```

`&.` guards a nil controller. It doesn't guard a controller that simply isn't an Inertia controller, which is the case that raises.

Why such controllers reach the middleware at all, from [`engine.rb:5-13`](https://github.com/inertiajs/inertia-rails/blob/e2d3881/lib/inertia_rails/engine.rb#L5-L13):

```ruby
app.middleware.use ::InertiaRails::Middleware          # :6  — unconditional, every app, every response

ActiveSupport.on_load(:action_controller_base) do      # :10 — Base only
end
```

So `ActionController::Base` descendants get `inertia_configuration` and `ActionController::API` descendants never do, while the middleware wraps both. In an API-first app most controllers are in the second group.

(The [`on_load(:action_controller)`](https://github.com/inertiajs/inertia-rails/blob/e2d3881/lib/inertia_rails/engine.rb#L15-L17) block just below registers the `:inertia` renderer for both Base and API, so it doesn't supply `inertia_configuration` either.)

That `:10` scoping came from [#77](https://github.com/inertiajs/inertia-rails/pull/77) ([`6f7a94f`](https://github.com/inertiajs/inertia-rails/commit/6f7a94f9db137cfcd0f1f2840f69af253f1a6f81)), which fixed this same class of crash (`undefined method 'helper' for ActionController::API`) at include time. The equivalent crash at request time is still there. Since `spec/dummy` has no `ActionController::API` controller, nothing covers the combination.

## Impact

Availability and noise, not data exposure or RCE:

- an unauthenticated caller can turn any non-Inertia endpoint into a 500 with one header, repeatably
- genuine 401/400 responses get masked as 500s
- error trackers fill with `NoMethodError` from anonymous traffic

None of that warrants an embargo, which is why I'm reporting it here rather than privately.

## Suggested fix

The real condition sits upstream of the crash: a request handled by a controller that can't speak the Inertia protocol isn't an Inertia request. Version staleness, force-refresh and 302→303 coercion have no meaning without an Inertia controller to serve them. Everything downstream funnels through [`inertia_request?`](https://github.com/inertiajs/inertia-rails/blob/e2d3881/lib/inertia_rails/middleware.rb#L94-L96), so guarding it there makes the protocol inert for those controllers and leaves `server_version` unreachable:

```ruby
# A controller that cannot speak the protocol is not an Inertia request,
# whatever the client claims in the header. nil (bare Rack endpoints,
# ActionCable) keeps today's behaviour.
def inertia_controller?
  handler = controller
  handler.nil? || handler.respond_to?(:inertia_configuration, true)
end

def inertia_request?
  @env['HTTP_X_INERTIA'].present? && inertia_controller?
end

# Defence in depth: guarantees the NoMethodError cannot return via a path
# that isn't gated by inertia_request?.
def server_version
  config = controller.send(:inertia_configuration) if inertia_controller? && !controller.nil?
  (config || InertiaRails.configuration).version
end
```

I went with `respond_to?(:inertia_configuration, true)` rather than `is_a?(ActionController::Base)` because responding to the method is the actual condition the crash site depends on; ancestry is only a proxy for it. It also holds up if you add an `on_load(:action_controller_api)` hook later, and it treats a controller that includes [`InertiaRails::Controller`](https://github.com/inertiajs/inertia-rails/blob/e2d3881/lib/inertia_rails/controller.rb) by hand as an Inertia controller.

Real Inertia controllers and nil controllers both fall through to today's behaviour.

On your suite the patch gives `833 examples, 7 failures`, identical to the unpatched baseline on `e2d3881` (the same 7 pre-existing `InstallGenerator` failures, unrelated).

## Guarding `server_version` alone isn't enough

The obvious patch, making `server_version` fall back to `InertiaRails.configuration`, removes the 500 but replaces it with a 409. It supplies a version to compare against, so a mismatched `X-Inertia-Version` now looks stale and the middleware answers with 409 + `X-Inertia-Location` via [`force_refresh`](https://github.com/inertiajs/inertia-rails/blob/e2d3881/lib/inertia_rails/middleware.rb#L110-L113). An anonymous caller can still steer a non-Inertia endpoint's response.

We shipped that half-fix ourselves first. Only a route-derived guard spec caught it.

It also passes a naive test, which is what makes it easy to miss. Measured against `spec/dummy` with the half-fix applied:

| server `version` | headers | result |
|---|---|---|
| `nil` (dummy default) | `X-Inertia` only | 200 — looks fixed |
| `nil` | `X-Inertia` + `X-Inertia-Version: deadbeef` | 409 |
| `'1.0'` | `X-Inertia` only | 409 |
| `'1.0'` | `X-Inertia` + `X-Inertia-Version: deadbeef` | 409 |

[`version_stale?`](https://github.com/inertiajs/inertia-rails/blob/e2d3881/lib/inertia_rails/middleware.rb#L98-L100) compares `coerce_version(client_version) != coerce_version(server_version)`, and both-nil is the only equal case. So `spec/dummy`'s default config plus a bare `X-Inertia` header is the one combination that hides the bug, and a test written that way goes green against the half-fix.

## Suggested regression spec

The config version and the version header matter as a pair: either one alone exposes the half-fix, but leave out both and the spec passes against it (the hiding cell in the table above). Both are set below.

```ruby
RSpec.describe 'X-Inertia on a non-Inertia controller', type: :request do
  with_inertia_config version: '1.0'   # load-bearing: dummy's nil default + no
                                       # version header is the one combination
                                       # that hides the bug

  it 'ignores the header on an ActionController::API endpoint' do
    get api_test_path
    baseline = response.status

    get api_test_path, headers: { 'X-Inertia' => 'true', 'X-Inertia-Version' => 'deadbeef' }

    expect(response.status).to eq(baseline)   # not 500, and not 409 either
  end

  it 'still 409s a stale Inertia GET on a real Inertia controller' do
    get empty_test_path, headers: { 'X-Inertia' => 'true', 'X-Inertia-Version' => 'stale' }
    expect(response.status).to eq 409
  end
end
```

Comparing each endpoint against itself rather than a hardcoded status matters here, since the 200-returning endpoints are most of the exposure and an "assert 401 stays 401" spec misses them.

Happy to open a PR with the patch and the spec (plus the `spec/dummy` `ActionController::API` controller) if that's useful.

SAY-5:
## Summary

Fixes #393.

A request carrying the `X-Inertia` header to a controller that isn't an Inertia controller (e.g. an `ActionController::API` action) raised `NoMethodError: undefined method 'inertia_configuration'` in the middleware, surfacing as a 500. The header alone is enough to trigger it, so any app with the gem installed is exposed regardless of whether it uses Inertia on that endpoint.

The middleware is installed on every request, but `InertiaRails::Controller` (and thus `inertia_configuration`) is only mixed into `ActionController::Base`. `server_version` called `controller&.send(:inertia_configuration)`, and the safe-navigation didn't help because the controller instance is present, just not Inertia-aware.

This gates `inertia_request?` on the controller actually responding to `inertia_configuration`, so a stray `X-Inertia` header on a non-Inertia endpoint no longer enters version negotiation. Real Inertia controllers are unchanged.

## Test plan

- Added a request spec covering an `ActionController::API` endpoint (asserted against its own baseline status) plus a guard that a stale Inertia GET on a real Inertia controller still returns 409.
- `bundle exec rspec spec/inertia/middleware_spec.rb` — 18 examples, 0 failures.
- Reverting only the middleware change reproduces the original `NoMethodError` in the new spec.

## Checklist

- [x] Tests added or updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-* Fix `NoMethodError` (500) when a request carrying the `X-Inertia` header reaches a non-Inertia controller such as an `ActionController::API` endpoint. The middleware now only applies version negotiation to controllers that include `InertiaRails::Controller` (@say)
+* Fix `NoMethodError` raised when a request carrying the `X-Inertia` header reaches a non-Inertia controller such as an `ActionController::API` endpoint (@SAY-5)
 
 ## [3.22.0] - 2026-07-17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+* Fix `NoMethodError` (500) when a request carrying the `X-Inertia` header reaches a non-Inertia controller such as an `ActionController::API` endpoint. The middleware now only applies version negotiation to controllers that include `InertiaRails::Controller` (@say)
+
 ## [3.22.0] - 2026-07-17
 
 * Only set the `XSRF-TOKEN` cookie on HTML and XHR responses. Other responses no longer carry a `Set-Cookie` that keeps CDNs from caching them — notably ActiveStorage's images, since its controllers also inherit from `ActionController::Base` (@acetinick)
--- a/lib/inertia_rails/middleware.rb
+++ b/lib/inertia_rails/middleware.rb
@@ -92,7 +92,15 @@ module InertiaRails
       end
 
       def inertia_request?
-        @env['HTTP_X_INERTIA'].present?
+        @env['HTTP_X_INERTIA'].present? && inertia_controller?
+      end
+
+      # The middleware is installed on every request, but Inertia's controller
+      # methods (e.g. inertia_configuration) are only mixed into Inertia-aware
+      # controllers. A stray X-Inertia header on a non-Inertia endpoint (such as
+      # an ActionController::API action) must not trigger version negotiation.
+      def inertia_controller?
+        controller.respond_to?(:inertia_configuration, true)
       end
 
       def version_stale?
--- a/lib/inertia_rails/middleware.rb
+++ b/lib/inertia_rails/middleware.rb
@@ -95,10 +95,6 @@ module InertiaRails
         @env['HTTP_X_INERTIA'].present? && inertia_controller?
       end
 
-      # The middleware is installed on every request, but Inertia's controller
-      # methods (e.g. inertia_configuration) are only mixed into Inertia-aware
-      # controllers. A stray X-Inertia header on a non-Inertia endpoint (such as
-      # an ActionController::API action) must not trigger version negotiation.
       def inertia_controller?
         controller.respond_to?(:inertia_configuration, true)
       end
--- a/spec/dummy/app/controllers/inertia_api_test_controller.rb
+++ b/spec/dummy/app/controllers/inertia_api_test_controller.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+# Inherits from ActionController::API, which does not receive the
+# InertiaRails::Controller mixin (only ActionController::Base does).
+class InertiaApiTestController < ActionController::API
+  def index
+    render json: { ok: true }
+  end
+end
--- a/spec/dummy/config/routes.rb
+++ b/spec/dummy/config/routes.rb
@@ -76,6 +76,7 @@ Rails.application.routes.draw do
   get 'cached_props' => 'inertia_render_test#cached_props'
   get 'cached_deferred_props' => 'inertia_render_test#cached_deferred_props'
   get 'non_inertiafied' => 'inertia_test#non_inertiafied'
+  get 'api_test' => 'inertia_api_test#index'
   get 'deeply_nested_props' => 'inertia_render_test#deeply_nested_props'
 
   get 'instance_props_test' => 'inertia_rails_mimic#instance_props_test'
--- a/spec/inertia/middleware_spec.rb
+++ b/spec/inertia/middleware_spec.rb
@@ -41,12 +41,6 @@ RSpec.describe 'InertiaRails::Middleware', type: :request do
 
       expect(response.status).to eq baseline
     end
-
-    it 'still refreshes a stale Inertia GET on a real Inertia controller' do
-      get empty_test_path, headers: { 'X-Inertia' => true, 'X-Inertia-Version' => 'stale' }
-
-      expect(response.status).to eq 409
-    end
   end
 
   context 'session loading guard' do
--- a/spec/inertia/middleware_spec.rb
+++ b/spec/inertia/middleware_spec.rb
@@ -30,6 +30,25 @@ RSpec.describe 'InertiaRails::Middleware', type: :request do
     end
   end
 
+  context 'X-Inertia header on a non-Inertia controller' do
+    with_inertia_config version: '1.0'
+
+    it 'ignores the header on an ActionController::API endpoint' do
+      get api_test_path
+      baseline = response.status
+
+      get api_test_path, headers: { 'X-Inertia' => true, 'X-Inertia-Version' => 'stale' }
+
+      expect(response.status).to eq baseline
+    end
+
+    it 'still refreshes a stale Inertia GET on a real Inertia controller' do
+      get empty_test_path, headers: { 'X-Inertia' => true, 'X-Inertia-Version' => 'stale' }
+
+      expect(response.status).to eq 409
+    end
+  end
+
   context 'session loading guard' do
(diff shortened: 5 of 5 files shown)


skryukov (approved):
Thanks @SAY-5!

PR #403, opened 2026-07-22
Skip the XSRF-TOKEN cookie under Rails 8.2's header-only forgery protection

mattwigham:
## Summary

Rails 8.2 introduces forgery protection verification strategies (rails/rails#56350): the new default, `:header_only`, verifies requests via the `Sec-Fetch-Site` header and never reads authenticity tokens. Under that strategy the `XSRF-TOKEN` cookie is pure overhead — a session write plus a cache-unfriendly `Set-Cookie` on every response, minting a token nothing will ever read.

This PR makes the adapter honor the strategy: when the responding controller verifies with `:header_only`, the `after_action` skips the cookie entirely, and expires a cookie left over from a token-based era so clients stop echoing it back as `X-XSRF-TOKEN`. This is the follow-through on the note in #368 about a "no cookie" mode becoming relevant once `:header_only` ships.

## Design

- **Feature-detected, inert on released Rails.** The strategy accessor only exists on `rails/main` today (milestoned 8.2.0), so the new branch is guarded with `respond_to?(:forgery_protection_verification_strategy)` — a no-op on every Rails version in the CI matrix, in the same 8.2-readiness spirit as #311.
- **Read per controller, not globally.** The strategy is read off the responding controller instance, so per-controller `protect_from_forgery using:` overrides are honored — an app can be header-only globally while one legacy surface keeps its cookie, and vice versa.
- **`:header_or_legacy_token` keeps today's behavior.** That strategy still falls back to token verification for browsers that don't send `Sec-Fetch-Site`, so it still needs the cookie (and the `xsrf_cookie_refresh` policies still govern it).
- **No client change needed.** Inertia's HTTP clients attach `X-XSRF-TOKEN` only when the cookie exists, so the exchange disappears cleanly from both sides.
- **Middleware untouched.** `copy_xsrf_to_csrf!` runs before dispatch (the controller strategy isn't knowable there), is harmless under `:header_only` (Rails never reads the header), and is still needed for the legacy strategy.
- The stale-cookie expiry emits a deletion `Set-Cookie` only when the request actually carried the cookie, so steady-state responses stay cookie-free.

## Test plan

- New specs in `spec/inertia/request_spec.rb` cover: no cookie under `:header_only`, deletion `Set-Cookie` for a stale cookie, and the cookie still being set under `:header_or_legacy_token`. The dummy controllers define `forgery_protection_verification_strategy` manually, so the behavior is exercised on the full released-Rails CI matrix without requiring edge Rails.
- Full RSpec suite, Minitest, and RuboCop green locally against Rails 8.1 (the only failures are the two Puma SSR process specs that fail identically on clean master).
- Verified end-to-end in an app running Rails `main` with `config.action_controller.forgery_protection_verification_strategy = :header_only` — i.e. the real config-delegate accessor rather than the spec stubs: no `XSRF-TOKEN` cookie is issued, a pre-existing cookie is expired once, and requests verify via `Sec-Fetch-Site`.

## Docs

- New "Header-Only CSRF Protection (Rails 8.2+)" section in the CSRF guide (`@available_since rails=master`).
- Notes in the HTTP-caching cookbook and the `xsrf_cookie_refresh` configuration docs that the refresh policies only apply to token-based strategies.

## Checklist

- [x] Tests added or updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+* Skip the `XSRF-TOKEN` cookie entirely (and expire a stale one) on controllers using Rails 8.2's `:header_only` forgery protection strategy, which verifies requests via `Sec-Fetch-Site` and never reads authenticity tokens (@mattwigham)
 * Add `config.server_head` to serialize meta tags as HTML strings into the `head` prop for the `serverHead` option of `createInertiaApp` (Inertia.js v3.5+), replacing the client-side cookbook component (@skryukov)
 * Add `meta_title_template` configuration option — a callable applied to the `<title>` tag of server driven meta tags; it receives the current title (or `nil`) and can provide a default for pages without one (@skryukov)
 * Add `ActiveSupport::Notifications` instrumentation: `render.inertia_rails`, `resolve_props.inertia_rails`, and `ssr.inertia_rails` events (@skryukov)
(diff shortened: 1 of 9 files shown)


skryukov (approved):
Thanks @mattwigham!

PR #404, opened 2026-07-28
Add Inertia DevTools recorder support

chasegiunta:
## Summary

Adds server-side Inertia DevTools recorder support to the Rails adapter.

The Chrome DevTools extension is backend-agnostic, but request inspection depends on each server adapter implementing the recorder protocol. Previously, Rails apps did not emit `X-Inertia-Devtools-Id` or expose `/_inertia/devtools/entries/:id`, so the panel could open without showing requests.

This PR implements the Rails side of the current protocol. No changes to `inertiajs/inertia-devtools` are required.

## What changed

- Records request entries in a bounded, expiring, thread-safe in-memory store.
- Emits `X-Inertia-Devtools-Id` and `X-Inertia-Devtools-Parent-Out`.
- Injects `script[data-inertia-devtools-id]` into initial full-page Inertia HTML responses.
- Exposes show and filtered-list endpoints under `/_inertia/devtools/entries`.
- Captures correlation headers, request type, timing, final response status, request/response headers and bodies, page props, Inertia prop metadata, and Rails route/controller metadata.
- Redacts common credential headers and payload keys.
- Deliberately omits streamed, binary, unserializable, and oversized response bodies.

## Safety

- Recording and reads are development-only by default.
- A configurable request authorization hook gates both recording and entry reads.
- Entries default to a 500-entry limit and one-hour TTL.
- DevTools endpoint requests bypass application routing and are never recorded.
- Recorder failures are isolated from normal application responses.

## Validation

- `bundle exec rake`
  - 930 RSpec examples, 0 failures
  - 52 Minitest runs / 98 assertions, 0 failures
  - 175 files inspected by RuboCop, no offenses

The full task was run with Node 24 because the current installer test toolchain requires Node `^20.19.0 || >=22.12.0`.

### DIFF
--- a/docs/guide/configuration.md
+++ b/docs/guide/configuration.md
@@ -93,6 +93,23 @@ end
 When enabled, props will be deep merged with shared data, combining hashes
 with the same keys instead of replacing them.
 
+### Inertia DevTools
+
+Inertia Rails includes the server-side recorder used by the Inertia DevTools browser extension. It is enabled and authorized only in the Rails development environment by default. Recorded entries are held in the current process only, capped at 500 entries, and expire after one hour.
+
+```ruby
+InertiaRails.configure do |config|
+  config.devtools_enabled = Rails.env.development?
+  config.devtools_max_entries = 500
+  config.devtools_entry_ttl = 1.hour
+  config.devtools_authorize = ->(request) { Rails.env.development? }
+end
+```
+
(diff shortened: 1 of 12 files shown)


chasegiunta:
> [!WARNING]
> PR was generated by 5.6 Sol by a non-Rails dev 😅, but I can confirm that it does work.

image

chasegiunta:
Superseded by https://github.com/inertiajs/inertia-rails/pull/405!

PR #408, opened 2026-08-29
Add scroll_props and once_props testing helpers

onk:
## Summary

`InertiaRails::Testing::TestResponse` only expose `deferred_props`, even though `PropsResolver#build_metadata` also makes other prop metadata.

scroll_props and once_props each have their own guide page ([infinite-scroll](https://inertia-rails.dev/guide/infinite-scroll), [once-props](https://inertia-rails.dev/guide/once-props)), so they are probably used a lot, and we want to test them too.

They are hash-shaped, like props/view_data/flash, so they can use the same test helpers already built for it, keeping this change small.

## Test plan

- Added RSpec (`spec/inertia/rspec_helper_spec.rb`) and Minitest (`test/integration/inertia_test.rb`) coverage for direct access and all matcher/assertion variants

## Checklist

- [x] Tests added or updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+* Add testing helpers for scroll props and once props (@onk)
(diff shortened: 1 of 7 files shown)


skryukov (approved):
Thanks @onk!

